PR #64, opened 2014-07-16
Use Node's EventEmitter

joseph-onsip:
This is very much a work in progress and isn't intended to be merged as-is. I'd just like to get the concept out here to see what we think of it. Here's a high-level summary:

`SIP.EventEmitter` now has the same API as [Node's EventEmitter](http://nodejs.org/api/events.html). `initEvents` and `initMoreEvents` are also defined, but they simply add a listener that uses the emitter's `logger` (no longer provided by `EventEmitter`, by the way) to log out a message when the event is emitted. The notion of an invalid event is no longer enforced.


wakamoleguy:
#66 has been merged.

PR #76, opened 2014-07-23
Media Handler: renderHint local/remote is an (Array of) HTMLMediaElement

joseph-onsip:
This removes one layer of nesting from the `invite`/`accept` `mediaHint`.

Note that `<video>` elements no longer get muted if an `<audio>` is provided alongside.

Usage (compare with 4096365):

```javascript
  var session = new SIP.UA().invite('welcome@onsip.com', {
    media: {
      render: {
        remote: [
          document.createElement('audio'), // included to demo array support
          document.createElement('video')
        ],
        local: document.createElement('video')
      }
    }
  });
```

### DIFF
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -67,9 +67,7 @@ Utils= {
             video: options.tagName === 'VIDEO'
           },
           render: {
-            remote: {
-              video: options
-            }
+            remote: options
           }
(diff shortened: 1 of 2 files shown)


wakamoleguy:
Merged.

PR #142, opened 2015-01-20
Fixes a typo in a function declaration (desugarSessionOptions).

wpp:
Just noticed this because it looked "weird" in my editor.

### DIFF
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -58,7 +58,7 @@ Utils= {
     options[winner] = options[winner] || options[loser] || defaultValue;
   },
 
-  desugarSessionOptions: function desugarSessionOptions (options) {
+  desugarSessionOptions: function (options) {
     if (global.HTMLMediaElement && options instanceof global.HTMLMediaElement) {
       options = {
         media: {


joseph-onsip:
Which editor do you use? It's strange that it would make this look "weird"...

For context, I generally include names in my function expressions so that they are easily identifiable in a stack trace.

wpp:
Viewed it in Sublime. I guess I just noticed it because it was the only function defined that way.
screen shot 2015-01-20 at 18 13 53

wpp:
I didn't realize that it's included in the stack-trace that way. Feel free to close it. I was just working on another PR and noticed this.

joseph-onsip:
Ah, yeah, there are some inconsistencies in the code, which should be ironed out one way or the other.

I looked a bit more into the stack-trace/call-stack issue and [found](https://stackoverflow.com/questions/15336347/why-using-named-function-expression) that browsers are smart enough to assign names to anonymous functions based on their context, so we could consider removing names in cases like this while preserving the usefulness of the stack.

However, in cases where the function is recursive, having a name lets us avoid using [arguments.callee](http://devdocs.io/javascript/functions/arguments/callee), so I'm leaning towards named function expressions being the standard style.

Issue #132, opened 2014-12-04
Calls erroring out due to headers stacking on subsequent calls


kylemeier:
Hi all,

I'm able to successfully make an initial call. However, if I hang up then call again, some headers sent by InviteClientContext() are now shown twice. Specifically, the headers generated by these lines (starting on line 6130 in sip.js (SIP version 0.6.4)):

  extraHeaders.push('Contact: '+ this.contact);
  extraHeaders.push('Allow: '+ SIP.Utils.getAllowedMethods(ua));
  if (!this.inviteWithoutSdp) {
    extraHeaders.push('Content-Type: application/sdp');
  } else if (this.renderbody) {
    extraHeaders.push('Content-Type: ' + this.rendertype);
    extraHeaders.push('Content-Disposition: render;handling=optional');
  }

Trace:
INVITE sip:tristan0636@testdomain.com SIP/2.0
Via: SIP/2.0/WS m3u4kcs4n0i7.invalid;branch=z9hG4bK645963
Max-Forwards: 70
To: sip:tristan0636@testdomain.com
From: sip:kyle1712@testdomain.com;tag=g7frq1u844
Call-ID: 3vrnkr0k0e3d9buu7m4i
CSeq: 3280 INVITE
Proxy-Authorization: Digest algorithm=MD5, username="53676338", realm="testdomain.com", nonce="VH+kIlR/ovYNUVH8FomCKJQcSsBMlbKT", uri="sip:tristan0636@testdomain.com", response="e0b535988e8531aa2eca044aa84bfd38"
Contact: sip:kyle1712@testdomain.com;gr=urn:uuid:0cca54cf-1840-4976-b7c5-6cd7e4010fda
Allow: ACK,CANCEL,BYE,OPTIONS,INFO,NOTIFY,INVITE
Content-Type: application/sdp
Contact: sip:kyle1712@testdomain.com;gr=urn:uuid:0cca54cf-1840-4976-b7c5-6cd7e4010fda
Allow: ACK,CANCEL,BYE,OPTIONS,INFO,NOTIFY,INVITE
Content-Type: application/sdp
Supported: gruu,outbound
User-Agent: SIP.js/0.6.4
Content-Length: 3589

This causes the call to immediately error out with a '400 Bad Content-Type Header'.

If I then call a third time, the headers are included three times in the invite. The headers continue to stack until I refresh the page.

Let me know if you need any additional information and any help would be greatly appreciated, thanks!

PR #143, opened 2015-01-20
Invite options

wpp:
I ran into the following situation today:
After a call to UA.invite:

    userAgent.invite("xyz@test.org", myopts);

The myopts object had a `extraHeaders` properties added.
(Session L1600).

The problem is if you have a setup like this:

      var myopts = {
        media: {
          constraints: {
            audio: true,
            video: true
          },
          render: {
            remote: {
              video: document.getElementById("remoteVideo")
            },
            local: {
              video: document.getElementById("localVideo")
            }
          }
        }
      };

    document.getElementById("callButton").onclick = function() {
      session = userAgent.invite("xyz@test.org", myopts);
      console.log(myopts);
    };

where you re-use the myopts object, you keep `push`ing extraHeaders
`Contact:` and `Allow:` and finally sending these with your invites.
I thought that a copy of the user-submitted object might be the cleanest
solution to this problem. But of course I'm open to other suggestions (and
alternative implementations of the copy functionality).

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "test": "grunt travis --verbose"
   },
   "dependencies": {
+    "extend": "^2.0.0",
     "pegjs": "0.8.0",
     "ws": "^0.6.4"
   },
--- a/src/Session.js
+++ b/src/Session.js
@@ -1,6 +1,7 @@
 module.exports = function (SIP, environment) {
 
 var DTMF = require('./Session/DTMF')(SIP);
+var extend = require('extend');
 
 var Session, InviteServerContext, InviteClientContext,
  C = {
@@ -1001,6 +1002,8 @@ InviteServerContext = function(ua, request) {
 
   this.logger = ua.getLogger('sip.inviteservercontext', this.id);
 
+  this.extend = extend;
+
   //Save the session into the ua sessions collection.
   this.ua.sessions[this.id] = this;
 
@@ -1270,7 +1273,7 @@ InviteServerContext.prototype = {
    * @param {Object} [options.media] gets passed to SIP.MediaHandler.getDescription as mediaHint
    */
   accept: function(options) {
-    options = options || {};
+    options = this.extend(true, {}, options) || {};
     options = SIP.Utils.desugarSessionOptions(options);
     SIP.Utils.optionsOverride(options, 'media', 'mediaConstraints', true, this.logger, this.ua.configuration.media);
     this.mediaHint = options.media;
--- a/src/UA.js
+++ b/src/UA.js
@@ -256,7 +256,7 @@ UA.prototype.afterConnected = function afterConnected (callback) {
  *
  */
 UA.prototype.invite = function(target, options) {
-  options = SIP.Utils.copy(options) || {};
+  options = options || {};
   options = SIP.Utils.desugarSessionOptions(options);
   SIP.Utils.optionsOverride(options, 'media', 'mediaConstraints', true, this.logger);
 
--- a/src/UA.js
+++ b/src/UA.js
@@ -256,7 +256,7 @@ UA.prototype.afterConnected = function afterConnected (callback) {
  *
  */
 UA.prototype.invite = function(target, options) {
-  options = options || {};
+  options = SIP.Utils.copy(options) || {};
   options = SIP.Utils.desugarSessionOptions(options);
   SIP.Utils.optionsOverride(options, 'media', 'mediaConstraints', true, this.logger);
 
--- a/src/UA.js
+++ b/src/UA.js
@@ -8,6 +8,9 @@
  * @param {Object} [configuration.media] gets passed to SIP.MediaHandler.getDescription as mediaHint
  */
 module.exports = function (SIP) {
+
+var extend = require('extend');
+
 var UA,
   C = {
     // UA status codes
@@ -78,6 +81,8 @@ UA = function(configuration) {
 
   this.log = new SIP.LoggerFactory();
   this.logger = this.getLogger('sip.ua');
+
+  this.extend = extend;
 
   this.cache = {
     credentials: {}
@@ -256,7 +261,7 @@ UA.prototype.afterConnected = function afterConnected (callback) {
  *
  */
 UA.prototype.invite = function(target, options) {
-  options = options || {};
+  options = this.extend(true, {}, options) || {};
   options = SIP.Utils.desugarSessionOptions(options);
   SIP.Utils.optionsOverride(options, 'media', 'mediaConstraints', true, this.logger);
 
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -92,6 +92,17 @@ Utils= {
     return 'v=0\r\n' + body.slice(start, end) + '\r\ns=-\r\nt=0 0\r\nc=IN IP4 0.0.0.0';
   },
 
+  copy: function(source) {
+    var copy = {}, prop;
+
+    for (prop in source) {
+      if (hasOwnProperty.call(source, prop)) {
+        copy[prop] = source[prop];
+      }
+    }
+    return copy;
+  },
+
   isFunction: function(fn) {
     if (fn !== undefined) {
       return Object.prototype.toString.call(fn) === '[object Function]';
--- a/src/Utils.js
+++ b/src/Utils.js
@@ -92,17 +92,6 @@ Utils= {
     return 'v=0\r\n' + body.slice(start, end) + '\r\ns=-\r\nt=0 0\r\nc=IN IP4 0.0.0.0';
   },
 
-  copy: function(source) {
-    var copy = {}, prop;
-
-    for (prop in source) {
-      if (hasOwnProperty.call(source, prop)) {
-        copy[prop] = source[prop];
-      }
-    }
-    return copy;
-  },
-
   isFunction: function(fn) {
     if (fn !== undefined) {
       return Object.prototype.toString.call(fn) === '[object Function]';
--- a/test/spec/SpecUA.js
+++ b/test/spec/SpecUA.js
@@ -430,7 +430,6 @@ describe('UA', function() {
 
       var options = {};
       UA.configuration.mediaHandlerFactory = function(){};
-      SIP.Utils.copy = function(){ return options; };
       UA.invite(target,options);
       // invite() puts the mediaHandlerFactory into the options object
       expect(SIP.InviteClientContext).toHaveBeenCalledWith(UA, target, options);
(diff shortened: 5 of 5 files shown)


wpp:
I'm not super happy with how the test-case was changed. If you decide that the test-case should actually test that the user-submitted options object should not be modified I'll add another, proper test case. But that might just have to wait till tomorrow. I'm going to bed :)

wakamoleguy:
Thanks for your pull request!  This particular issue is a bit of a controversial subject.  You can read a little about it on this thread, along with the previously suggested workaround:

https://github.com/onsip/SIP.js/issues/132

What that means is that we may just take a little more convincing to merge this in, as I'm not convinced that it fully solves the issue and an easy workaround is in place for most cases.

On a non-blocking note, I personally don't like the SIP.Utils object and would like to deprecate it eventually in favor of smaller, more descriptive modules.  We are a ways from actually doing that, but I am not terribly enthused about adding a method to it which is only used in one or two places.  That wouldn't actually stop us from accepting a pull request, though...just airing my thoughts.

Anyways, if you read that Issue and still think that this is the way to go on the request, I'll take a closer look at both the change and the test case, and we'll see what we can do.

Cheers,

-Will

wpp:
Hi Will,

Thanks for the quick feedback. My apologies for not searching the issues before submitting the PR, I don't want to waste your time!

If I understand it correctly, @kylemeier ended up with a workaround like:

```javascript
  document.getElementById("callButton").onclick = function() {
      var myoptions = {
          media: {
            constraints: {
              audio: true,
              video: true
            },
            render: {
              remote: {
                video: document.getElementById("remoteVideo")
              },
              local: {
                video: document.getElementById("localVideo")
              }
            }
          }
      };
      session = userAgent.invite("xyz.com", myoptions);
    };
```
The problem I (personally) have with this approach is if
I also want to listen for invites later on and wanted to use the same `myoptions`:

```javascript
  userAgent.on("invite", function(session) {
    if (confirm('Incoming call...')) {
      session.accept(myoptions);
    } else {
      session.reject({statusCode: 603, reasonPhrase: "Decline"});
    }
  });
```
The first thing I'd do is extract the rather verbose `myoptions` so it can be shared between the two functions. Which would result in the same issue. (This is how I encountered the problem.)

The other workaround @joseph-onsip proposed:

```javascript
var ua = new SIP.UA({traceSip: true});
var opts = function () {
    return {};
};

var session = ua.invite('chickens@junctionnetworks.com', opts());
session.on('accepted', function () {
    session.bye();
    ua.invite('chickens@junctionnetworks.com', opts())
});
```

Would solve this problem.

**But** I'd still have to remember why I  wrapped this in a function if I come back to the code 6 months later, [when the documentation has a "normal" object](http://sipjs.com/guides/make-call/).

In our case it was not easy to detect the problem in the first place, because we didn't have `traceSip` set to `true` and even when we did it wasn't trivial to spot in the console. At the end of the day *I just didn't expect my submitted options to be mutated by the library.*

[Upon further investigation](http://stackoverflow.com/questions/728360/most-elegant-way-to-clone-a-javascript-object) I think that the copy implementation I submitted probably needs to be checked for more edge cases. I'm also very open to any other approaches (not copying the object) and more discussions.

TLDR: At the end of the day *I* would love (to help) to have a non-workaround based solution to this issue. And I would rather invest more of my time to finding a "good" solution so that other people don't run into this.

Sorry for being such a nitpicker :smile: ! And thanks again for a the quick feedback (as always).
Cheers, Philipp

wpp:
> Cloning the options object is not an option, as there are properties on it which cannot be cloned (such as MediaElements). That leaves us with creating a white-list of acceptable options, also an unpalatable choice given that a UA or Session does not necessary know what options the MediaHandler, for example, can accept.

Shit. If a mediaStream is part of the `options` there is basically no way of cloning them.

wakamoleguy:
> At the end of the day I just didn't expect my submitted options to be mutated by the library.
I think this is the crux of it.  Really, we *shouldn't* be mutating the options objects passed in.  The problem there is that it is so engrained into the library that it would be very tedious to fix.  I could point fingers about the origin of that practice, but the truth is that we as authors haven't really made any effort to stop doing it.

The primary reason to modify them at all is to set default options when none are provided.  At one point, @joseph-onsip suggested doing something clever with JavaScript prototypes, but I was able to poke holes in early attempts at that.  I think there is a way to do this properly, where we can promise not to touch the object.  It may just involve pushing "default" behavior down lower into the library, or passing it in a separate object.

joseph-onsip:
> I think there is a way to do this properly

After taking another look at the invite/accept code, I agree. Instead of adding its own stuff to the user options, sip.js should just produce a fresh default options object for each invite/accept, then recursively [extend](https://www.npmjs.com/package/extend) it with the user options.

wpp:
@joseph-onsip So would you rather have defaults before-hand and extend them with the user options along the lines of:

```javascript
UA.prototype.invite = function(target, userOptions) {
    options = extend(true, defaultOptions, userOptions) || {};
    ...
```

...or extend on an empty object and add the options afterwards (which would require less code changes I guess)

```javascript
UA.prototype.invite = function(target, userOptions) {
    defaultOptions = extend(true, {}, userOptions) || {};
    ...
```

>  I'm thinking we should just let that (continue to) be the case and document it.
So in the case where a mediaStream is submitted, the user options would still be mutated?

```javascript
UA.prototype.invite = function(target, userOptions) {
    if (isCloneable(userOptions)) {
      defaultOptions = extend(true, {}, userOptions) || {};
    ...
````

[I guess the mediaStream is already a bit of a "nuanced" case](http://sipjs.com/guides/reuse-mediastreams/) and if the documentation would reflect that it'd be ok?

If you already have an implementation in mind please feel free to go for it. I'm not insisting that this PR will eventually get merged. But if you don't have the time and can wait till the weekend I'll be happy to provide another take at this.

Cheers

**Comment by joseph-onsip** [comment]
> ...or extend on an empty object and add the options afterwards (which would require less code changes I guess)

> ```javascript
UA.prototype.invite = function(target, userOptions) {
    defaultOptions = extend(true, {}, userOptions) || {};
    ...
```

That's what I had in mind, yes.

> So in the case where a mediaStream is submitted, the user options would still be mutated?

No, the options object would remain the same, but since `extend` copies the stream by reference, any changes made by SIP.js to the stream would be reflected in the user options. Here's a (Chrome-only) [demo](http://requirebin.com/?gist=7dbf9f564b7d0de99623) (open the console)

> I guess the mediaStream is already a bit of a "nuanced" case and if the documentation would reflect that it'd be ok?

Yes, the stream is part of the `media` property of the invite/accept options, which in theory could be an arbitrary object (depending on the MediaHandler implementation being used) and is therefore not guaranteed to be cloneable. I'd definitely be comfortable with making this distinction explicit.

> But if you don't have the time and can wait till the weekend I'll be happy to provide another take at this.

I took a brief shot at this last night, but ran into issues with the testing. I may give it another go, but I can't guarantee anything personally.

wakamoleguy:
I'm personally stuck working on the termination events for a little while, and so I won't be able to get to this just yet.  Something like those extensions sound like they should work, probably.

wpp:
This is a little embarrassing but I'm not familiar with npm don't know how to properly add https://www.npmjs.com/package/extend as a dependency. I added it under `"dependencies"` in `package.json` and `$ npm installed` but still get undefined errors when running grunt tasks.

joseph-onsip:
You should be able to install `extend` and add it to package.json by running:

```shell
npm install --save extend
```

To actually use it, you'll first need to add this to the top of the file you're calling it from:

```javascript
var extend = require('extend');
```

wpp:
Sorry for taking so long to reply. I gave this another shot **but**:

I'm not sure how you'd like `extend` to be "included" in the objects (UA and Session).
I tried to see how other `require`s are used (e.g. Session `var DTMF = require('./Session/DTMF')(SIP);`).

Just let me know how (and if) I should change it. I'm not so convinced that this is worth bothering you guys again. Sorry for that.

But if you think this is fine I'll also make a PR for the documentation to address the fact that `mediaStream`s could still be modified.

> I guess the mediaStream is already a bit of a "nuanced" case and if the documentation would reflect that it'd be ok?

wakamoleguy:
I'm going to take a look at this from the ground up.  I'd prefer to get this in before the next version release.

wpp:
Great! I'll leave this PR open for now but feel free to close whenever you want.

wakamoleguy:
I spent some time fleshing out a plan of action.  Well, kind of.  It is more of a list of constraints to abide by as we tackle this.  "You" is used to refer mostly to the file/module at hand.

1. UA passes options straight through.
1a. Syntactic sugar is handled on the Context constructor side.
2. Move syntactic sugar from Utils into proper context.
3. Options that get passed down through the Context to other constructors must be whitelisted or name-spaced.
3b. Context must be aware of and control any namespace related syntactic sugar (like Media options)
3c. Options that get passed down should really be new objects.
4. You must understand options that you handle.
5. Defaults explicitly defined.
5b. List defaults in the function or top of the file (should be easy to find and update for both coder and peruser.)
5c. (Also, because of 4) Clone your own options object, inheriting from the defaults.
5d. It is not necessary for defaults to exist (if the option is optional), but it should be seen as bad practice.
6. Move options overrides out of Util.
6a. If it happens in one place, move it there.
6b. If it happens in multiple places, make an Options.js module.
7. Document all options.
8. Verify that the above practices recurse down through *all* files.

I'm generally leaning towards ignoring the "MediaStream cloning" issue.  I've thought a lot about it, and I think there is a subtle distinction between options objects that we should not modify, and values for those options which are being given to SIP.js with the express purpose of us mutating them.  So far the MediaStream is the only thing I can think of that does that, but there may be others along the way.

wakamoleguy:
Open question:  If users aren't expecting SIP.js to change the options object, do they expect to be able to modify it themselves?  How does them mucking with the original object affect asynchronous behavior?

wakamoleguy:
Proposed answer:
* Parameters that are used directly should synchronously be copied somewhere safe.  `this.body = options.body`
* If we aren't going to use the options directly, we don't know enough about them to copy them.  We should save the options away as-is, but not go deeper than we understand.  `this.mediaHint = options.mediaHint` but not `this.constraints = options.mediaHint.constraints`.  If the user mangles those sub-options before we pass them along, so be it.  Once the MediaHandler (in this case) actually needs them, it will copy them to a safe place itself in whatever state they happen to be in at the time.
* Potentially requires a reworking, but in that example above, if the Session or something needs to re-use the same mediaHint, it should ask its MediaHandler for the safely copied version.

wakamoleguy:
I've merged in a branch that tackles the main options cleanup.  I'm going to close this ticket for now, since most of the rest of the goals can be tackled through either other tickets (like Utils cleanup) as well as generic documentation updates and avoiding doing this certain ways in the future.

PR #153, opened 2015-02-25
Makes options to RegisterContext's register "instance" properties.

wpp:
For example:

```javascript
userAgent = new SIP.UA({
  ...
  traceSip: true,
  register: false
});

userAgent.register({traceSip: false, extraHeaders: [ 'X-Foo: foo', 'X-Bar: bar' ] });
```

`RegisterContext` receives the options via

`register({traceSip: false, extraHeaders: [ 'X-Foo: foo', 'X-Bar: bar' ] })`

before the connection to the socket is established. 

Once the transport is connected
`RegisterContext's.onTransportConnected` is called and registers with empty
options.

This commit changes the options so that the RegisterContext saves them
as 'instance properties', which allows for calling onTransportConnected
with the register options.

Although it feels a little weird calling: `this.register(this.options);`
I like the way the options provided to `userAgent.register()` will always overwrite
the instance properties. 

But as always please feel free to critique and make suggestions!
Hope I didn't miss anything. Noticed this while working on another PR.

### DIFF
--- a/src/RegisterContext.js
+++ b/src/RegisterContext.js
@@ -48,13 +48,13 @@ RegisterContext = function (ua) {
 
 RegisterContext.prototype = {
   register: function (options) {
-    var self = this;
+    var self = this, extraHeaders;
 
     // Handle Options
     this.options = options || {};
-    this.extraHeaders = (this.options.extraHeaders || []).slice();
-    this.extraHeaders.push('Contact: ' + this.contact + ';expires=' + this.expires);
-    this.extraHeaders.push('Allow: ' + SIP.Utils.getAllowedMethods(this.ua));
+    extraHeaders = (this.options.extraHeaders || []).slice();
+    extraHeaders.push('Contact: ' + this.contact + ';expires=' + this.expires);
+    extraHeaders.push('Allow: ' + SIP.Utils.getAllowedMethods(this.ua));
 
     this.receiveResponse = function(response) {
       var contact, expires,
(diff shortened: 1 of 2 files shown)


wakamoleguy:
Hi Philipp,
I like this pull request.  It is definitely odd that `onTransportConnected` doesn't allow for any options to be specified.  I just have a few concerns before pulling it in.

1. Should the RegisterContext's `close` method also call `unregister` with `this.options`?
2. I would prefer that `extraHeaders` not be cached on `this`.  It doesn't seem necessary to accomplish what this pull request sets out to do, and `extraHeaders` tend to be modified in places further down the line that could pollute the cached array.  Cloning it from `this.options` each time feels safer to me.

With those changes addressed (or with comments telling me why I'm utterly wrong, which is not unheard of), I'd be happy to pull this in.

Also, you get bonus points for fixing the whitespace errors. :-)

Cheers,

-Will

wpp:
Hi Will,

Thanks for the feedback.

1. I can't come up with a scenario where that might be an issue:
`userAgent.stop()` is the only place where `RegisterContext.close()`
`(and therefore .unregister())` could be called.
So for the options to unregister `{all: true, extraHeaders: [ 'X-Foo: foo', 'X-Bar: bar' ]` to get lost,
we would have to call `.unregister(options)` and only afterwards have `userAgent.stop()`.
```javascript
userAgent.unregister({all: true, extraHeaders: [ 'X-Foo: foo', 'X-Bar: bar' ]});
userAgent.stop();
```
And at that point the options would be ignored anyway. (`RegisterContext.unregister()`)
```javascript
if(!this.registered) {
  this.logger.warn('already unregistered');
  return;
}
```
But in case I might miss something and that could happen: The options to `unregister` differ to those provided to `register`. So calling unregister with the `this.options` would not be desired anyway. A solution would be to save `this.registerOptions` and `this.unregisterOptions`. Which I would gladly provide in case that is needed.
2. Yep, I agree just pushed another commit.

> Also, you get bonus points for fixing the whitespace errors. :-)

I'll forever cherish these bonus points.

PR #155, opened 2015-02-26
Example implementation for current contact list

wpp:
Hi,

I really **don't intend to have this merged** but I thought it might be easier to illustrate this with code.

I'm looking for a way to make a ["Request for Current Contact List" rfc3665](https://tools.ietf.org/html/rfc3665#section-2.3)

> Bob sends a register request to the Proxy Server **containing no
   Contact headers**, indicating the user wishes to query the server for
   the user's current contact list.  Since the user already has
   authenticated with the server, the user supplies authentication
   credentials with the request and is not challenged by the server.
   The SIP server validates the user's credentials.  The server returns
   a response (200 OK) which includes the user's current registration
   list in Contact headers.

I would like to be able to register a userAgent depending on the current contact list.
If I haven't missed anything there is currently no way to send a register without a contact header.

This PR and the code below ("client code")
shows how I hacked around this for now.

```javascript

userAgent.register({noContact: true});

userAgent.registerContext.on('no-contact', function () {
      userAgent.register();
    });

    userAgent.registerContext.on('contacts', function (contactList) {
      for (var idx = 0; idx < contactList.length; idx++) {
        if (contactList[idx].indexOf(instanceID) > -1) {
          instanceIdfree = false;
        } else {
          instanceIdfree = true;
        }
      }
```

I would love to have this functionality in SIP.js but I'm not sure how to tackle this and how the API would look like.
I.e. should perhaps something like:

```javascript
userAgent.getCurrentContactList()
```

If you point me in the right direction and don't have time to work on this I'd be happy to give it a shot.

### DIFF
--- a/src/RegisterContext.js
+++ b/src/RegisterContext.js
@@ -8,7 +8,9 @@ RegisterContext = function (ua) {
       regId = 1,
       events = [
         'registered',
-        'unregistered'
+        'unregistered',
+        'no-contact',
+        'contacts'
       ];
 
   this.registrar = ua.configuration.registrarServer;
@@ -53,7 +55,9 @@ RegisterContext.prototype = {
     // Handle Options
     this.options = options || {};
     extraHeaders = (this.options.extraHeaders || []).slice();
-    extraHeaders.push('Contact: ' + this.contact + ';expires=' + this.expires);
+    if (!this.options.hasOwnProperty('noContact')) {
+      extraHeaders.push('Contact: ' + this.contact + ';expires=' + this.expires);
+    }
     extraHeaders.push('Allow: ' + SIP.Utils.getAllowedMethods(this.ua));
 
     this.receiveResponse = function(response) {
@@ -90,6 +94,7 @@ RegisterContext.prototype = {
 
           // Search the Contact pointing to us and update the expires value accordingly.
           if (!contacts) {
+            his.emit('no-contact');
             this.logger.warn('no Contact header in response to REGISTER, response ignored');
             break;
           }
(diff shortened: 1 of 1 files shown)


wakamoleguy:
I commented on the mailing list as well.  I am not familiar with this spec from the get-go, so I will have to do some reading to understand all of the implications.

On the pull request itself, I like the idea of passing an option to the RegisterContext and reusing some of that code.  It keeps the REGISTER method code isolated in one place, which is good.  I'm not sure whether or not this deserves its own method on the UA, or if it is enough to just pass it as an option to `ua.register`.  That may depend on how it plays with the current register/re-register loop.  Also, I think the events could be either combined into one (which maybe passes an array of contact headers, which may be empty?) or combined with an existing event (like `accepted`, where the contacts are available on the response object).  In the end, this just becomes a question of how much sugar to pour on it.

To phrase that in terms that could give you direction, in case you want to go off on your own:
* As a first cut, I'd provide it as an option to `UA.register()`.  That should make the request, without interfering with the current registration status.  The return value should be the context, which will emit ClientContext events normally.  The contact list can be found by the application using the `accepted` event of that context.
* As a second cut, we can decide if we want to add that as a new UA API method (and what to call it) and if we should create the special `contacts` event (and what specific arguments to pass).

Two last quick notes:
* I'd prefer that you check the existence of the option without using hasOwnProperty.  If somebody passes a fancy inherting options object, we should respect their prototypes and whatnot.
* The tests currently pass once you fix [this typo](https://github.com/onsip/SIP.js/pull/155/files#diff-cc6af5ce58cdd20707dd4dd1398be049R97)

wakamoleguy:
Practically speaking, while the RegisterContext is the nicest place for this, you may have better luck making a one-off ClientContext and specifying the REGISTER method.

wpp:
@wakamoleguy fyi: I don't really need this functionality any longer. If I'm **not** the only one who has brought it up, feel free top reopen this or poke me for an implementation. I'll  close this for now so you don't have open PR's lingering around forever.

PR #164, opened 2015-04-02
Improve the WebSocket error reporting

capilkey:
If your encounter a WebSocket connection problem the current error message reports as:
"Thu Apr 02 2015 16:14:18 GMT-0500 (CDT) | sip.transport | WebSocket connection error: [object Event]" sip.js:2636

If you instead stringify the error object before appending it you will get more useful information.

### DIFF
--- a/src/Transport.js
+++ b/src/Transport.js
@@ -265,7 +265,7 @@ Transport.prototype = {
   * @param {event} e
   */
   onError: function(e) {
-    this.logger.warn('WebSocket connection error: ' + e);
+    this.logger.warn('WebSocket connection error: ' + JSON.stringify(e));
   },
 
   /**


wakamoleguy:
Thank you for your pull request!  It looks like native error objects should not contain circular references, so this looks fine as is.

shaneneuerburg:
I'm running into an issue with this PR and React Native. The JSON.stringify throws an error about a circular reference when the error occurs. In particular, it happens when the SIP server goes offline/is inaccessible. For a temporary workaround, I set up a try-catch around that error handling, but obviously the last thing you want to do in an "onError" method is try to handle another error :). However, I'm struggling to propose a better idea.

egreenmachine:
Can you provide some more information? Specifically, I would like to know what the `e` object looks like.

shaneneuerburg:
Here is a screenshot from the inspector:

https://photos.app.goo.gl/Rlt3pCME2BcNIEFS2

shaneneuerburg:
Thank you!

Issue #19, opened 2014-05-19
SIP.js tries to update in-dialog SDP on the Peer Connection


egreenmachine:
If SIP.js receives an in-dialog Invite with new SDP it tries to update the Peer Connection. In Chrome 34 this causes the Peer Connection to crash. 

```
Mon May 19 2014 11:47:02 GMT-0400 (EDT) | sip.invitecontext.mediahandler | peerConnection.setLocalDescription failed sip-0.5.0.js:168
Mon May 19 2014 11:47:02 GMT-0400 (EDT) | sip.invitecontext.mediahandler | Failed to set local answer sdp: Session error code: ERROR_CONTENT. Session error description: Failed to set audio receive codecs.. 
```

I think that we should handle this case before SIP.js tries to set new SDP on the Peer Connection so that it does not try to update the SDP.

PR #170, opened 2015-05-13
Re-enable receiving reINVITEs

joseph-onsip:
Browsers have gotten better at renegotiation (now that [Firefox 38 is out](https://hacks.mozilla.org/2015/03/webrtc-in-firefox-38-multistream-and-renegotiation/)), so let's consider turning this back on.

### DIFF
--- a/src/Session.js
+++ b/src/Session.js
@@ -663,9 +663,7 @@ Session.prototype = {
       case SIP.C.INVITE:
         if(this.status === C.STATUS_CONFIRMED) {
           this.logger.log('re-INVITE received');
-          // Switch these two lines to try re-INVITEs:
-          //this.receiveReinvite(request);
-          request.reply(488, null, ['Warning: 399 sipjs "Cannot update media description"']);
+          this.receiveReinvite(request);
         }
         break;
       case SIP.C.INFO:
--- a/test/spec/SpecSession.js
+++ b/test/spec/SpecSession.js
@@ -1595,7 +1595,7 @@ describe('InviteServerContext', function() {
     });
 
     describe('method is INVITE', function() {
-      xit('calls receiveReinvite', function() {
+      it('calls receiveReinvite', function() {
         InviteServerContext.status = 12;
         req = SIP.Parser.parseMessage([
           'INVITE sip:gled5gsn@hk95bautgaa7.invalid;transport=ws;aor=james%40onsnip.onsip.com SIP/2.0',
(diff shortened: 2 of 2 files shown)


joseph-onsip:
In my local checkout, I merged this into 873c9da186ce6699db15a0ee4c06e254b9b52cf6 (master branch at time of writing) and built a [test page](http://jsfiddle.net/q0hb9gpa/7/) for trying out hold/unhold (and therefore receiveReinvite) with different combinations of browsers/audio/video/inviteWithoutSdp. I'll see if I can break anything, and update this with my findings.

joseph-onsip:
Ok, after a bunch of testing, I found that there's still a couple of problems with cross-browser WebRTC. #182 is one such problem, and another is as follows: If Firefox makes an audio call with SDP to Chrome, and Chrome answers with audio then tries to `hold()`, Firefox refuses to renegotiate printing the following error:

> DOMException [InvalidSessionDescriptionError: "ICE restart is unsupported at this time (new remote description changes either the ice-ufrag or ice-pwd)ice-ufrag (old): XNaL6MUvAo6+hfTmice-ufrag (new): mQDDAkDOY9Yf9wJoice-pwd (old): 5MmxCArSUOdddCA7HQgOE5+6ice-pwd (new): WX9lUGO1SxoGP3hu0h4szV+V"
code: 0
nsresult: 0x0]

However, Firefox does send a 488 in this case, and the call carries on as usual (note that `hold()` also disabled the audio MediaStreamTrack, but `unhold()` will re-enable it), instead of with a "crashed PeerConnection" as in #19.

Given that making this change allows the browsers to at least try renegotiating (and gracefully [falling back to sending either a 488 or a 500 if it fails](https://github.com/joseph-onsip/SIP.js/blob/bd5c3b21db50c33036c9b229b1581ea7a327d45e/src/Session.js#L594-L603)), I think this is fine to merge in.

wpp:
Hi @joseph-onsip,

glad you're working on this! Hopefully I've got some time this week to test this as well and perhaps help you out a little. I'd be looking into switching/adding/removing streams on existing `PeerConnection`s. Looks like your most recent work is on master as of now?

BTW: AFAIK cross-browser renegotiation [is at least a couple of months out](https://code.google.com/p/chromium/issues/detail?id=465349) since Chrome and FF use different implementations of multistream ([the link you posted for FF 38](https://hacks.mozilla.org/2015/03/webrtc-in-firefox-38-multistream-and-renegotiation/)):

> Also, it is important to note that Google Chrome’s current implementation of multistream is not going to be interoperable; this is because Chrome has not yet implemented the specification for multistream (called “unified plan” – check on their progress in the Google Chromium Bug tracker). Instead they are still using an older Google proposal (called “plan B”). These two approaches are mutually incompatible.

But in case someone is looking for a temporary work-around there is a [polyfill](https://www.npmjs.com/package/sdp-interop) which might help out.

PR #173, opened 2015-05-19
Unwrap terminated data

kylemeier:
The terminated event currently wraps its callback data in an object,
making it inconsistent with the docs (as well as how the other events
operate). This change simply unwraps it.


wakamoleguy:
Good catch.  Thank you!

PR #177, opened 2015-05-22
Allow uppercase host

ryan-rowland:
It doesn't seem like the rfc3261 spec requires host to be lowercase, and in some cases this is limiting.

Relevant specifications can be found at https://www.ietf.org/rfc/rfc3261.txt on page 222.


ryan-rowland:
I will re-open this request when I have relevant examples, and have fixed the affected tests.

PR #178, opened 2015-05-26
Prevent simultaneous UA starts

ryan-rowland:
Running into an issue where multiple calls to UA.start() before the UA has connected will result in multiple open web sockets. It's possible to prevent this in our implementation, but this seems like logic that belongs on the UA, and other SIPjs users may benefit from.


wakamoleguy:
Thanks!

-Will

PR #179, opened 2015-05-26
Fix reregister options

ryan-rowland:
We're inside an anonymous function so `this` doesn't refer to the `RegisterContext`. Changed to `self` so that options are properly re-sent on re-register.


joseph-onsip:
Nice catch, thanks for the fix!

PR #180, opened 2015-05-26
Pass options to unregister on close

ryan-rowland:
We've run into a problem where we need to pass authorization headers to both register and unregister. This is easy to handle with register and unregister directly, but close doesn't seem to utilize our stored register options. This PR simply changes the `RegisterContext.close` function to pass `this.options` to `this.unregister()`.

### DIFF
--- a/src/RegisterContext.js
+++ b/src/RegisterContext.js
@@ -192,7 +192,7 @@ RegisterContext.prototype = {
 
   close: function() {
     this.registered_before = this.registered;
-    this.unregister();
+    this.unregister(this.options);
   },
 
   unregister: function(options) {
--- a/test/spec/SpecRegisterContext.js
+++ b/test/spec/SpecRegisterContext.js
@@ -159,10 +159,11 @@ describe('RegisterContext', function() {
       expect(RegisterContext.registered).toBe(RegisterContext.registered_before);
     });
     
-    it('calls unregister', function() {
+    it('calls unregister with RegisterContext\'s stored options', function() {
       expect(RegisterContext.unregister).not.toHaveBeenCalled();
+      RegisterContext.options = { foo: 'bar' };
       RegisterContext.close();
-      expect(RegisterContext.unregister).toHaveBeenCalledWith();
+      expect(RegisterContext.unregister).toHaveBeenCalledWith(RegisterContext.options);
     });
   });
   


wakamoleguy:
Hi Ryan,

First of all, thank you for your contributions.  This one in particular is interesting, and we've discussed it previously in Pull Request #153.  Since `this.unregister()` actually takes different options than `register()`, we don't want to reuse those options blindly.  As an easy workaround, you can unregister explicitly before calling `stop()`.  If you'd like to convince me otherwise, I'm certainly open to it.

Cheers,

-Will

ryan-rowland:
Hey Will,

Thanks for looking at my PR!

I may be missing something, but it looks like the options sent to `RegisterContext.register` and `RegisterContext.unregister` are the same (`extraHeaders`) except unregister has an optional `all` setting. It seems that calling `RegisterContext.close` currently calls `RegisterContext.unregister` without any options, so it always runs with `{ all: false }`. This should be the case with the method implemented in this PR, unless the end user explicitly adds `all: true` to his register function.

If consistency is preferred, we can clone `this.options` and override `all` to `false` before passing them. This effectively ensures a valid set of unregister options (At least to the degree that register and unregister themselves ensure a valid set of options) is passed.

Again, maybe I'm missing something. But it seems to me that the only option `RegisterContext` cares about is `extraHeaders` and for unregister, `all`. Maybe an alternative would be to store an `extraHeaders` property rather than the entire `options` hash, since that's all it's used for? Maybe then it would make more sense to re-use it for `close()`.

I do understand that an end-user can manually call:

```
this.unregister(options);
this.close();
```

But should they have to? They have already supplied their authorization headers, and `RegisterContext` has already stored them. This also means that instead of relying on the built-in cleanup on window close (in `UA`: `environment.addEventListener('unload', this.stop.bind(this));`) they will have to handle it in their own event handler. And I haven't tested this, but if they don't remove UA's handler, might there even then be a race condition?

Just my two cents. :grinning:

wakamoleguy:
So, with the current set of options, it would be okay.  I'm more concerned about adding future options.  If we make this change, any option we add to `register` has to also work (or be ignored appropriately) for `unregister`.  It doesn't seem very future proof, especially if there is an easy workaround.

> I do understand that an end-user can manually call...But should they have to?

I think so, if we are trading the flexibility of the library for one line of code that not a clear majority of users need.

> ...might there even then be a race condition?

Race condition between what?  Sending the `REGISTER` request and closing the WebSocket?  `close` would call `unregister` synchronously, so calling it earlier isn't going to cause any race with the connection that wouldn't already be there.  The only other thing I can think of is if calling `unregister` twice in a row races, which is a separate issue that we would want to patch.

> Maybe an alternative would be to store an extraHeaders property rather than the entire options hash, since that's all it's used for?

That's an interesting idea.  If the goal is to reuse specific headers on all registration requests, perhaps we could carry that over.  I'd still want to reference it as a property of the RegistrationContext, rather than reusing the options.  Do we know if extraHeaders would *always* want to be used in `close`, or is that something we should make configurable?

ryan-rowland:
> Race condition between what?
If UA's `on('unload')` fires before mine, `stop` would be called before `unregister`. I haven't tested this, but I wanted to bring it up as a possible concern.

> Do we know if extraHeaders would always want to be used in close, or is that something we should make configurable?

I can't speak for all use-cases, but in ours (authorization headers) we always want to be passing them. I support the idea of making it configurable, if there are supporting use-cases and it's done intuitively.

I could envision something similar to:
```
register: function(options) {
  this.closeHeaders = options.closeWithHeaders ? options.extraHeaders : [];
  // ...
},

close: function() {
  var options = {
    all: false,
    extraHeaders: this.closeHeaders
  };
  this.unregister(options);
}
```

What do you think?

ryan-rowland:
@wakamoleguy I can code this up for review and update any associated tests if this looks good to you. Just let me know if you've got a better idea in mind so I can hopefully knock it out in one pass. :grinning:

wakamoleguy:
> If UA's on('unload') fires before mine, stop would be called before unregister. I haven't tested this, but I wanted to bring it up as a possible concern.

Oh, right.  I had forgotten that we added that.  Fair point.

> I can code this up for review and update any associated tests if this looks good to you.

I think what you've outlined looks good.  Just please make sure that the extraHeaders gets copied into its own array properly (we already do this elsewhere in the file), to avoid messing with people who reuse the same options object for different calls.

Thank you for volunteering, and for putting up with my nitpickiness.

PR #181, opened 2015-05-29
Added RegisterContext.closeHeaders logic

ryan-rowland:
Added support for a `closeWithHeaders` flag on `RegisterContext.register`. If true, the supplied headers will be sent to `RegisterContext.unregister` from `RegisterContext.close`. Related discussion can be found at #180


wakamoleguy:
Looks good.  Thanks!

PR #188, opened 2015-06-15
Allow uppercase host

ryan-rowland:
Is there a specific reason that hosts are automatically lower-cased?

Our SIP addresses follow this scheme:

```
sip://clientName@accountID.foobar.com
```

Where clientName is the client's name as registered with SIP and accountID is the user's account. An account can have multiple named clients.

However, our accountIDs are case-sensitive and contain capital letters. Is there a reason case-insensitivity is being enforced on this level or might these be a case of over-fitting?

### DIFF
--- a/src/Grammar/src/Grammar.pegjs
+++ b/src/Grammar/src/Grammar.pegjs
@@ -130,7 +130,7 @@ password        = ( unreserved / escaped / "&" / "=" / "+" / "$" / "," )* {
 hostport        = host ( ":" port )?
 
 host            = ( hostname / IPv4address / IPv6reference ) {
-                    options.data.host = text().toLowerCase();
+                    options.data.host = text();
                     return options.data.host; }
 
 hostname        = ( domainlabel "." )* toplabel  "." ? {
--- a/test/spec/SpecGrammar.js
+++ b/test/spec/SpecGrammar.js
@@ -60,7 +60,7 @@ describe('Grammar', function () {
 
         uriHas('scheme', 'sip');
         uriHas('user', '+1234');
-        uriHas('host', 'aliax.net');
+        uriHas('host', 'ALIAX.net');
         uriHas('port', undefined);
 
         uriReturns('getParam', 'transport', 'ws');
@@ -78,7 +78,7 @@ describe('Grammar', function () {
         c1.setParam('New-Param', null);
         expect(c1.hasParam('NEW-param')).toEqual(true);
         c1.uri.setParam('New-Param', null);
-        expect(c1.toString()).toEqual('"€€€" <sip:+999@aliax.net;transport=ws;new-param>;+sip.instance="zxCV";new-param');
+        expect(c1.toString()).toEqual('"€€€" <sip:+999@ALIAX.net;transport=ws;new-param>;+sip.instance="zxCV";new-param');
       });
     });
 
@@ -96,7 +96,7 @@ describe('Grammar', function () {
       c2Has('displayName', undefined);
 
       c2Returns('hasParam', 'HEADERPARAM', true);
-      c2Returns('toString', null, '<sip:bob@biloxi.com>;headerparam');
+      c2Returns('toString', null, '<sip:bob@biloxi.COM>;headerparam');
 
       describe('its URI', function () {
         var uriThunk = function () { return c2.uri; };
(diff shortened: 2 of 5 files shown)


joseph-onsip:
It looks like the lowercasing was introduced [quite a while back](https://github.com/onsip/SIP.js/commit/7b345bfff0ce87d9f2266cb5516959b04df23ad4#diff-d20dca32b077f94d242847b0cf3c02d6R129) to help compare URIs more correctly [per RFC 3261](http://tools.ietf.org/html/rfc3261#section-19.1.4):

>       o  Comparison of the userinfo of SIP and SIPS URIs is case-
         sensitive.  This includes userinfo containing passwords or
         formatted as telephone-subscribers.  Comparison of all other
         components of the URI is case-insensitive unless explicitly
         defined otherwise.
> ...
>    The URIs within each of the following sets are equivalent:

>    sip:%61lice@atlanta.com;transport=TCP
>    sip:alice@AtLanTa.CoM;Transport=tcp

Given the above, it sounds like your accountIDs should *not* be case-sensitive. Granted, URIs could be compared without changing their internal representations or serializations, but that would likely involve more changes to the code.

ryan-rowland:
@joseph-onsip Could you point me to the URI comparison method you're referring to, that would need to be changed to accommodate this change?

I understand that you're enforcing RFC 3261, and I believe sticking to specifications is a good thing. However, I would argue that enforcing this limitation does not serve to bring SIP.js closer to spec, but rather attempts to force the user to follow spec. If that's the goal, understood. I personally believe it would be appropriate to give users the freedom to regard specifications as they see fit outside the confines of the internal mechanisms of SIP.js.

In our case, we have a large enough user base and complex enough of a platform that it would be infeasible to change all of the account IDs. We currently have a fork with this change implemented, however one of our recent goals has been to give back to SIP.js. We feel that in most cases, if we're having issue, there's probably at least one other party having the same. If we can devise a fix or configuration option for ourselves, there's a good chance it will be helpful to someone else.

etamme:
He linked directly to the comparison method in RFC3261, as well as copy pasting the relevant paragraph which explicitly says that userinfo is case sensitive, and that all other portions of the URI are case insensitive.

http://tools.ietf.org/html/rfc3261#section-19.1.4

SIP.js functionality by default should absolutely be as compliant with the core RFC3261 as possible.  If anything, I would suggest making a configuration option to turn on case sensitivity for the host portion of the URI.  Simply removing the case insensitivity is not a good idea as it could very likely change how SIP.js interops with other SIP user agents that ARE following the spec.

wakamoleguy:
The specific case I was thinking of was RURI matching in the UA, to determine if a request is for that UA.  It turns out that the code for that, though, [only looks at the user portion](https://github.com/onsip/SIP.js/blob/master/src/UA.js#L586).  There may be other places like this where we (or our users) are looking to see if two URIs are equivalent, and the case matters.  I think that's the primary reason why we are being so cautious here.  Changing the type of data that gets into the internals of the library could cause subtle issues in the long run.

I am not philosophically opposed to this change (which is saying something...I was when I started looking at this).  It makes sense that somebody should be able to specify a representation of a valid SIP URI and have that representation carried forward.  That being said, we currently use an equivalent representation (by RFC standards) and we know it works internally.

I want to find a way for this to work such that it changes existing behavior a minimal amount.  I would see this as a change to the URI structure.  Basically, I'd like to keep the existing `uri.host` and `uri.toString` behavior (lowercase) while providing the raw, original URI for those who need it.  Two caveats would be:

1. What if the URI was not created via parsing?  I'd suggest populating `.raw` or `.toRaw` with the `toString` representation.
2. Would sending the raw URI out on the wire instead of the toLowerCase string cause any problems with other proxies or UAs?

markandrus:
> It makes sense that somebody should be able to specify a representation of a valid SIP URI and have that representation carried forward. That being said, we currently use an equivalent representation (by RFC standards) and we know it works internally.

I agree. Normalizing the host portion to lowercase simplifies comparison internally, but it would be nice if the original representation was sent out over the wire.

I think 1. is reasonable and 2. should be fine since, per the RFC, comparison of the host portion is case-insensitive. Other proxies and UAs should follow this.

ryan-rowland:
As this implementation is largely different than what was submitted in this original PR, I have opened a new PR here: https://github.com/onsip/SIP.js/pull/191

PR #189, opened 2015-06-18
Override receiveResponse for .refer()

markandrus:
I'd like to use `refer` for the "Requesting a Focus to Add a New Resource to a Conference" use case described in [RFC 4579, Section 5.5](https://tools.ietf.org/html/rfc4579#section-5.5). Would you be open to allowing developers to override SIP.js's existing behavior in `receiveResponse`?

Also, I found that a 200 OK would not be sent to the NOTIFY unless explicitly handled.

### DIFF
--- a/src/Session.js
+++ b/src/Session.js
@@ -236,7 +236,7 @@ Session.prototype = {
     this.sendRequest(SIP.C.REFER, {
       extraHeaders: extraHeaders,
       body: options.body,
-      receiveResponse: function (response) {
+      receiveResponse: options.receiveResponse || function (response) {
         if ( ! /^2[0-9]{2}$/.test(response.status_code) ) {
           return;
         }
@@ -720,6 +720,9 @@ Session.prototype = {
             this.emit('refer', request);
           }
         }
+        break;
+      case SIP.C.NOTIFY:
+        request.reply(200, 'OK');
         break;
     }
   },


wakamoleguy:
That seems like a use case that we should be able to support.  I'm not completely sold on overriding the receiveRequest method entirely, though.  Is the main issue that the session is terminated, or that there is no visibility into the actual response?  I'm guessing the former, since state updates tend to come from future NOTIFY requests.

Let me think on this a little.  It could be OK, but I want to give careful consideration to adding new options, especially since adding it to `refer` would create a precedent for other methods.

markandrus:
> Is the main issue that the session is terminated, or that there is no visibility into the actual response?

I'm actually handling both: I circumvent the session termination and use the status code of the response to resolve or reject a Promise. I plan to also consume the NOTIFYs, but I'm not yet doing so.

I agree this deserves some consideration since it would create a precedent for other methods.

wakamoleguy:
Thanks for the feedback.  I'm still not sure what to do.  I have jury duty for a few days, which should offer ample time for mulling.

On a side note...you can totally disregard this if you'd like, but I came across this post the other day (@joseph-onsip actually linked me to it).  It covers one of my worst habits in using Promises, and your comment reminded me of it:  https://github.com/petkaantonov/bluebird/wiki/Promise-anti-patterns#the-thensuccess-fail-anti-pattern  In short, using the 'catch' functionality to do control flow can cause pain later on.  Handling expected failures (that do not throw exceptions) within the `then` is considered better practice.  (I'm half saying this just to remind myself, because I have a whole project where that is the primary way of doing things.)

markandrus:
@wakamoleguy Perhaps another path forward could be to make `Session.sendRequest` part of the public (albeit lower-level) API?

wakamoleguy:
Whoops, sorry for letting this fall through the cracks.  Thank you for keeping tabs on it.

> Perhaps another path forward could be to make Session.sendRequest part of the public (albeit lower-level) API?

I think this is the way to go.  I already consider `sendRequest` to be public, as it is the default way right now to send an in-dialog request with a custom method (similar to `UA.prototype.request`).  Since the side receiving the refer can override its default behavior (by simply not calling `followRefer`), that also sounds like it would work for your case.

So, if that's the case, I'll add it to our documentation.  I'm surprised it isn't already there.

markandrus:
@wakamoleguy Thanks for the update! I changed my code to use `sendRequest` and it works fine, so I'm happy with this solution (moving it into the public API also helps me with some other use cases where I need to send in-dialog requests).

However, I still have to pass in `extraHeaders`—a few of which I have to construct using `session.ua.normalizeTarget(target)` and `SIP.utils.getAllowedMethods(session.ua)`, which I'm not excited about. Any ideas on this?

markandrus:
@wakamoleguy I can confirm that without the

```js
case SIP.C.NOTIFY:
  request.reply(200, 'OK');
  break;
```

code added by this PR, SIP.js will not reply to NOTIFY. Would you be open to merging a separate PR with this case added?

wakamoleguy:
Yeah, I'd definitely take a pull request with just the NOTIFY behavior.  I'm surprised we haven't seen that before.

I would just merge this request and then change it, but then syncing up the commit message and author, etc... A pull request would be cleanest.

PR #190, opened 2015-06-19
Expand keep-alive support

markandrus:
SIP.js recognizes [keep-alive responses](https://github.com/onsip/SIP.js/blob/608407e6153e7a65c5c48c2f140fb4a943e4b5d6/src/Transport.js#L202). What do you think about enabling keep-alive requests?

We could expose this through a UA configuration option, `keepAliveInterval`, to set the upper bound for sending double-CRLF keep-alives as described in [RFC 5626, Section 4.4.1](https://tools.ietf.org/html/rfc5626#section-4.4.1). `keepAliveInterval` defaults to zero, which disables sending keep-alives.

Open questions:

- Detecting the `Flow-Timer` header and updating the `keepAliveInterval`
- Detecting when a pong has not been received for 10 seconds

### DIFF
--- a/src/Transport.js
+++ b/src/Transport.js
@@ -30,6 +30,9 @@ Transport = function(ua, server) {
   this.reconnectTimer = null;
   this.lastTransportError = {};
 
+  this.keepAliveInterval = ua.configuration.keepAliveInterval;
+  this.keepAliveTimer = null;
+
   this.ua.transport = this;
 
   // Connect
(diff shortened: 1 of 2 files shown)


joseph-onsip:
After looking at various parts of RFC 5626, I think this is definitely a step in the right direction. As you mentioned, there are things which have yet to be sorted out in order to properly implement the RFC, but I don't see anything wrong with offering the configuration option, especially since the default value doesn't yet cause anything to be sent.

However, I don't think `computeKeepAliveTimeout` works as intended, after trying it out a few times. It should be something like this instead:

```javascript
function computeKeepAliveTimeout (upperBound) {
  var lowerBound = upperBound * 0.8;
  return 1000 * (Math.random() * (upperBound - lowerBound) + lowerBound);
}
```

markandrus:
@joseph-onsip Good catch on `computeKeepAliveTimeout`, thanks! Fixed in dac3c42.

ryan-rowland:
@joseph-onsip Friendly bump here; currently working on implementing support for treating the flow as failed when the keepAlive response time exceeds 10 seconds, per [RFC 5626, Section 4.4.1](https://tools.ietf.org/html/rfc5626#section-4.4.1), which will depend on the final implementation of this PR. :grinning:

joseph-onsip:
Thanks @markandrus for the contribution and @ryan-rowland for the bump. Glad to see more full-fledged keep-alive support on the way!

markandrus:
Thanks, @joseph-onsip!

PR #191, opened 2015-06-19
Use raw address URIs over SIP

ryan-rowland:
Implementation of outcome discussed on PR https://github.com/onsip/SIP.js/pull/188.

 * I moved the lowercasing out of `Grammar.pegjs` and into the `URI` constructor, so that it has access to the original string.
 * I created a `toRaw()` method that will create a string using the original (raw) values and plugged it in where appropriate. `toString()` is unaffected.
 * Added tests for the new functionality.


ryan-rowland:
@wakamoleguy Is this in line with what we discussed in https://github.com/onsip/SIP.js/pull/188?

wakamoleguy:
I think so, although I admit that the conversation is not fresh in my mind at this point.  I'll review this pull request this week.  Thank you for your patience!

PR #195, opened 2015-07-01
Emit 'keepAliveTimeout' event when Keep-Alive times out

ryan-rowland:
Small step toward implementing RFC spec. For now, just emitting an event that we can catch when a Keep-Alive times out (After 10 seconds per [RFC 4.4.1](https://tools.ietf.org/html/rfc5626#section-4.4.1)). This emits 'keepAliveTimeout' from the UA.

### DIFF
--- a/src/Transport.js
+++ b/src/Transport.js
@@ -31,6 +31,7 @@ Transport = function(ua, server) {
   this.lastTransportError = {};
 
   this.keepAliveInterval = ua.configuration.keepAliveInterval;
+  this.keepAliveTimeout = null;
   this.keepAliveTimer = null;
 
   this.ua.transport = this;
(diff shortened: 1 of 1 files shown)


wakamoleguy:
This look solid.  I like that it just fires an event, so that the actual behavior of tearing down the flow is handled at an application level.  Different endpoints may or may not support it, but the event is flexible.

For those wondering, the 10 seconds is a hardcoded value recommended by [RFC5626](https://tools.ietf.org/html/rfc5626#page-20):

> Note: This value of 10-second timeout was selected to be long
>      enough that it allows plenty of time for a server to send a
>      response even if the server is temporarily busy with an
>      administrative activity.  At the same time, it was selected to be
>      small enough that a UA registered to two redundant servers with
>      unremarkable hardware uptime could still easily provide very high
>      levels of overall reliability.  Although some Internet protocols
>      are designed for round-trip times over 10 seconds, SIP for real-
>      time communications is not really usable in these type of
>      environments as users often abandon calls before waiting much more
>      than a few seconds.

PR #202, opened 2015-08-01
Timer B should be reset on provisional responses

benlangfeld:
When transitioning from CALLING to PROCEEDING and on every further provisional response, we should reset Timer B. Precedent is provided by Kamailio (http://kamailio.org/docs/modules/4.2.x/modules/tm.html#restart_fr_on_each_reply) and the use-case is for call queueing relying on 182 Queued status, which is periodically retransmitted while the call remains queued.

### DIFF
--- a/src/Transactions.js
+++ b/src/Transactions.js
@@ -172,13 +172,18 @@ InviteClientTransaction.prototype.stateChanged = function(state) {
 };
 
 InviteClientTransaction.prototype.send = function() {
-  var tr = this;
   this.stateChanged(C.STATUS_CALLING);
-  this.B = SIP.Timers.setTimeout(tr.timer_B.bind(tr), SIP.Timers.TIMER_B);
+  this.resetTimerB();
 
   if(!this.transport.send(this.request)) {
     this.onTransportError();
   }
+};
+
+InviteClientTransaction.prototype.resetTimerB = function() {
+  var tr = this;
+  SIP.Timers.clearTimeout(this.B);
+  this.B = SIP.Timers.setTimeout(tr.timer_B.bind(tr), SIP.Timers.TIMER_B);
 };
 
 InviteClientTransaction.prototype.onTransportError = function() {
@@ -280,12 +285,14 @@ InviteClientTransaction.prototype.receiveResponse = function(response) {
     switch(this.state) {
       case C.STATUS_CALLING:
         this.stateChanged(C.STATUS_PROCEEDING);
+        this.resetTimerB();
         this.request_sender.receiveResponse(response);
         if(this.cancel) {
           this.transport.send(this.cancel);
         }
         break;
       case C.STATUS_PROCEEDING:
+        this.resetTimerB();
         this.request_sender.receiveResponse(response);
         break;
     }


benlangfeld:
Ping @bklang for cursory review.

bklang:
:+1: based on my reading of the Kamailio module docs

etamme:
Rfc3261 says nothing about resetting timer b.  The tm module param you are referencing is related to "resetting on reply" which..
Doesn't make sense for a final response time out.  By definition it should fire unless a final response is recieved.

I'll be at ClueCon with another OnSIP engineer and we can discuss if you want to grab us.

benlangfeld:
Unlike @bklang, I unfortunately won't be at ClueCon.

Kamailio specifies a separate timer (http://kamailio.org/docs/modules/4.2.x/modules/tm.html#max_inv_lifetime, Timer C) which is an overall timer from INVITE to final response and not reset on provisional responses.

wakamoleguy:
Regardless of how Kamailio names its timers (which appears to be [inconsistent with RFC3261](https://tools.ietf.org/html/rfc3261#page-108)), using it as a 'precedent' for this type of behavior doesn't really make sense.  Kamailio is a SIP proxy, while SIP.js implements a SIP UA.  They have different behavior.

Reading the RFC, I reach the same conclusion as @etamme.  Timer B is behaving as it should, closing out a transaction *if and only if* the session has not reached the proceeding state.  If the session does reach the proceeding state, it should not timeout until a final response has been received.

This behavior should also behave appropriately in the case of repeated `182 Queued` responses, as since there is no timeout, the session will sit around waiting on queue.  In fact, Timer B will have no effect if there is even a single `182 Queued` response received, so resetting it seems unnecessary in that case, even if it we were to adopt the nonstandard behavior.  Proxies do have other responsibilities (like Timer C) to ensure that the transaction is healthy, but those are the responsibility of the proxy, as defined.

I could foresee two reasons why the `182 Queued` would be having issues right now:

* You are connecting SIP.js directly to a UAS, rather than to a SIP proxy that handles Timer C.  That's not really what the RFC would consider 'typical' (it describes an outbound proxy and the SIP trapezoid as such).  I would expect some peculiarities if you have that setup.  If the UAS does send a `182 Queued`, you'll still be OK.  The lack of timeouts doesn't seem to be a real issue, either, as you're guaranteed to have reached the target endpoint in one hop.
* The PeerConnection will time out after ~30 seconds anyways.  Regardless of the SIP timers you have set up, you're going to want to get that PeerConnection completed fast.  Using Invite w/o SDP or 100rel would get around this.

So, uh, can anybody point to a standard defining this behavior?  Otherwise, I'm going to close out the request.

benlangfeld:
> Reading the RFC, I reach the same conclusion as @etamme. Timer B is behaving as it should, closing out a transaction if and only if the session has not reached the proceeding state. If the session does reach the proceeding state, it should not timeout until a final response has been received.

That's exactly what this pull request implements. Prior to this change, Timer B expires after reaching the proceeding state (in my case, receiving 100 followed by 182). Logs to follow showing this. With this change, the call is allowed to sit in the proceeding state as long as it receives repeated provisional responses before Timer B expires. See also comments at http://www.ietf.org/mail-archive/web/sip/current/msg10305.html.

> This behavior should also behave appropriately in the case of repeated 182 Queued responses, as since there is no timeout, the session will sit around waiting on queue. In fact, Timer B will have no effect if there is even a single 182 Queued response received, so resetting it seems unnecessary in that case, even if it we were to adopt the nonstandard behavior. Proxies do have other responsibilities (like Timer C) to ensure that the transaction is healthy, but those are the responsibility of the proxy, as defined.

This is not the case prior to this change. After this change is is the case.

> I could foresee two reasons why the 182 Queued would be having issues right now:
>
> * You are connecting SIP.js directly to a UAS, rather than to a SIP proxy that handles Timer C. That's not really what the RFC would consider 'typical' (it describes an outbound proxy and the SIP trapezoid as such). I would expect some peculiarities if you have that setup. If the UAS does send a 182 Queued, you'll still be OK. The lack of timeouts doesn't seem to be a real issue, either, as you're guaranteed to have reached the target endpoint in one hop.

Nope, SIP.js is connected to Kamailio. The call is routed to Mobicents. Mobicents sends 182 every 28 seconds. Timer B still expires. This commit fixes my use case.

> * The PeerConnection will time out after ~30 seconds anyways. Regardless of the SIP timers you have set up, you're going to want to get that PeerConnection completed fast. Using Invite w/o SDP or 100rel would get around this.

I have not seen evidence of the PeerConnection timing out, and this commit brings my application into a working state.

> So, uh, can anybody point to a standard defining this behavior? Otherwise, I'm going to close out the request.

I cannot because there does not appear to be a SIP specification which defines it one way or the other. I do, however, believe I have made arguments that show that a UA which does *not* implement this is broken in a user-facing way. If I'm mistaken, I'd love to be corrected.

wakamoleguy:
> Prior to this change, Timer B expires after reaching the proceeding state (in my case, receiving 100 followed by 182).

Timer B *should* expire.  It just doesn't *do* anything if the session is no longer in the calling state.  I confirmed with a test call that if a 1xx response is received, even after Timer B expires, the call is not timed out.  As far as I can tell from the RFC and this discussion, this is precisely the desired behavior.

benlangfeld:
> Timer B should expire. It just doesn't do anything if the session is no longer in the calling state. I confirmed with a test call that if a 1xx response is received, even after Timer B expires, the call is not timed out. As far as I can tell from the RFC and this discussion, this is precisely the desired behavior.

You know, you're absolutely right. I rolled back my change, downgraded back to 0.6.2 as the application was before, and no `CANCEL` is sent as I previously observed. I guess I had been drinking on Friday or something :/ More likely, I think I probably got in a bad state with Kamailio configuration and in the bugfix fog failed to recognise that the SIP.js log statement `sip.transaction.ict | Timer B expired for INVITE client transaction z9hG4bK6229049` was actually harmless, despite being printed within a second of the `CANCEL`.

Thank you for helping me clear this up and understand better what was going on. Apologies for the noise.

wakamoleguy:
No apology necessary.  It's always good to keep us in check.  Plus, I personally came out of this with a better understanding of the timers and their behaviors in UAs and proxies.  I also apologize if I seemed overly aggressive at any point, which I guarantee was just my effort to be overly specific and focus on the standards.

In any case, we've talked at one point about changing the wording of those logs statements to say something other than 'expired.'  You're not the first one to assume that something is happening (especially something bad).

benlangfeld:
I'm not particularly concerned about the terminology "expired"; it's consistent with discussion of timers in many contexts. Had I seen what I saw when I rolled back today, I never would have had an issue. It was purely the interference of whatever was sending a `CANCEL` and my failure at the time to spot it.

Issue #18, opened 2014-05-19
Handle Session Timers


egreenmachine:
FreeSwitch by default has Session Timeouts in the SIP headers. This header is documented in [RFC4028](http://tools.ietf.org/html/rfc4028)  

Session-Expires: 120;refresher=uas

The re-invites from FreeSwitch have identical SDP to what it sent on the original 200 OK (answer to the INVITE). We should be able to handle it by just sending a 200 OK with the SDP from the original INVITE. No ports or IP addresses have changed.

One potential issue I could see with this is if FreeSwitch decides to update the SDP to use a different IP address or Port. SIP.js has no way of updating the Peer Connection

markandrus:
@joseph-onsip there are a few commits on your branch that it looks like it would be nice to have in master, namely

- [SIPMessage: dedupe Supported header generation into getSupportedHeader](https://github.com/joseph-onsip/SIP.js/commit/ef520533774ddc1c1a901552527ec2a13e6943a4)
- [Grammar: parse Session-Expires/Min-SE header into object](https://github.com/joseph-onsip/SIP.js/commit/daeb113db9f75c52025988c3e937842c0f334d2c)
- [Grammar: parse Supported/Require header into list of values](https://github.com/joseph-onsip/SIP.js/commit/ea0863700b16718f29c23d9ad11ee1bd4dc42aa0) (I think this one would be useful even before session timer support)

any chance of getting those merged in, even before full session timer support is ready?

joseph-onsip:
Hey @markandrus, sorry for the latency on this. I've created https://github.com/onsip/SIP.js/pull/203 on your behalf, so the changes can be considered there.

james-criscuolo:
Some work was done here, and is pushed on this branch:

https://github.com/onsip/SIP.js/tree/session-timers

The branch is a work in progress, but we hope it at least hits all the features we are looking to incorporate here.

digitaldan:
Any updates or workarounds for this?

PR #203, opened 2015-08-04
Header handling

joseph-onsip:
Requested by @markandrus in https://github.com/onsip/SIP.js/issues/18#issuecomment-124682335

These changes add parsing of the following headers:

* Session-Expires
* Min-SE
* Supported
* Require

They also clean up the serialization of outgoing Supported headers.


markandrus:
Thanks @joseph-onsip. I have cherry-picked some of these commits on a fork. They have been working fine for me, except they also needed your 5d291260b57e79c751c47f6f14a374c628ec1ba2 change. Sorry, I had not yet noticed this when commenting on #18.

The deduped `getSupportedHeader` function is particularly useful, as I would like to support setting additional option tags on `UA` construction and this looks like a good place to put it. So +1 from me if that counts :-)

joseph-onsip:
Thanks for verifying it works well in practice, I'll go ahead and merge.

PR #207, opened 2015-08-13
Allow setting additional Supported option tags

markandrus:
It would be nice to be able to add Supported option tags in addition to those already set by SIP.js. This PR adds a new configuration option, `supported`, taking an array of option tags to append to the Supported header.

Open questions:
- This is as simple as possible—just concat the Arrays—but it also means that you could duplicate option tags, including those that SIP.js already sets (path, gruu, 100rel, replaces, outbound). Is this something that should be protected against (e.g., by using a Set)?
- The name `supported` might suggest the configuration option overwrites all the option tags that _would be_ set in the Supported header. Would it be more appropriate to name it, e.g., `extraSupportedOptionTags`?

### DIFF
--- a/src/Constants.js
+++ b/src/Constants.js
@@ -156,6 +156,45 @@ return {
     603: 'Decline',
     604: 'Does Not Exist Anywhere',
     606: 'Not Acceptable'
+  },
+
+  /* SIP Option Tags
+   * DOC: http://www.iana.org/assignments/sip-parameters/sip-parameters.xhtml#sip-parameters-4
+   */
+  OPTION_TAGS: {
+    '100rel':                   true,  // RFC 3262
+    199:                        true,  // RFC 6228
+    answermode:                 true,  // RFC 5373
+    'early-session':            true,  // RFC 3959
+    eventlist:                  true,  // RFC 4662
+    explicitsub:                true,  // RFC-ietf-sipcore-refer-explicit-subscription-03
+    'from-change':              true,  // RFC 4916
+    'geolocation-http':         true,  // RFC 6442
+    'geolocation-sip':          true,  // RFC 6442
+    gin:                        true,  // RFC 6140
+    gruu:                       true,  // RFC 5627
+    histinfo:                   true,  // RFC 7044
+    ice:                        true,  // RFC 5768
+    join:                       true,  // RFC 3911
+    'multiple-refer':           true,  // RFC 5368
+    norefersub:                 true,  // RFC 4488
+    nosub:                      true,  // RFC-ietf-sipcore-refer-explicit-subscription-03
+    outbound:                   true,  // RFC 5626
+    path:                       true,  // RFC 3327
+    policy:                     true,  // RFC 6794
+    precondition:               true,  // RFC 3312
+    pref:                       true,  // RFC 3840
+    privacy:                    true,  // RFC 3323
+    'recipient-list-invite':    true,  // RFC 5366
+    'recipient-list-message':   true,  // RFC 5365
+    'recipient-list-subscribe': true,  // RFC 5367
+    replaces:                   true,  // RFC 3891
+    'resource-priority':        true,  // RFC 4412
+    'sdp-anat':                 true,  // RFC 4092
+    'sec-agree':                true,  // RFC 3329
+    tdialog:                    true,  // RFC 4538
+    timer:                      true,  // RFC 4028
+    uui:                        true   // RFC 7433
   }
 };
 };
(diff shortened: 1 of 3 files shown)


wakamoleguy:
A few initial thoughts:

* I'm not sure if duplicating options tags is okay or not.  RFC 3261 at least doesn't mention one way or another.  I would suspect that it is okay, but we should probably deduplicate them to be safe.
* I think it would be more appropriate to be prefixed with `extra`.  Even just `extraSupported` sounds okay to me (although maybe a little silly).  Do others have thoughts?

My biggest concern is more of a philosophical one. RFC 3261 explicitly states:

> The option tags listed MUST only refer to extensions defined in
>   standards-track RFCs.  This is to prevent servers from insisting that
>   clients implement non-standard, vendor-defined features in order to
>   receive service.  Extensions defined by experimental and
>   informational RFCs are explicitly excluded from usage with the
>   Supported header field in a request, since they too are often used to
>   document vendor-defined extensions.

I know we usually leave it to the application, and not supporting the RFC is their choice.  But in a situation where the RFC explicitly calls out custom vendor-prefixes... I'm not so sure we should be making it easier.  Such uses could still be accomplished through custom forks of SIP.js.  It's a hassle for applications doing that (which probably serves them right), and it would certainly clear my conscience of supporting further siloification of the SIP world.

But then again, I know there are a lot of standards-track-defined options tags that we don't support right now.  By not having this option, are we purposefully obstructing those valid uses?  Are all of those situations ones in which we could accept a pull request, or are there options that we'll never support out of the box?  I suspect that's the case.

I'm probably being too much of an idealist on this one.  Somebody please jump in.

james-onsip:
I think 'limiting good cases' is a much worse situation than 'allowing bad cases,' and the 'leave it to the application' argument we generally use is fine here. That said, we COULD (and I know this would be hard (impossible?)) have a list of `extensions defined in standards-track RFCs` and check the incoming supported arguments against that, rejecting any that are not standard. I think this pull request could be accepted without that, but it's something we could do to address your concern in the future.

Also, I agree with the first two bullet points. Maybe `userSupported` would be appropriate, seeing as our library is not advertising support outright?

markandrus:
@wakamoleguy 

> I'm not sure if duplicating options tags is okay or not. RFC 3261 at least doesn't mention one way or another. I would suspect that it is okay, but we should probably deduplicate them to be safe.

Agree. I couldn't find language confirming this, but I've also never seen any messages with duplicated option tags. I've added deduplication to my commit.

> I think it would be more appropriate to be prefixed with `extra`. Even just `extraSupported` sounds okay to me (although maybe a little silly). Do others have thoughts?

I like this and have updated my commit to reflect `extraSupported`. @james-onsip, I'd prefer `extraSupported` to `userSupported` simply because it matches `extraHeaders`, but I could swing either way.

In terms of philosophy, I think I understand your concern: I can't deny that I intend to use this feature to signal support for optional, non-standard functionality. However, I would like to see SIP.js expose interfaces powerful enough such that, following your example, an application could implement standard SIP extensions on top of SIP.js (perhaps in advance of SIP.js landing proper support, or in the case where SIP.js support is unlikely).

wakamoleguy:
Hmm.  I may come around eventually, but I would definitely accept James's idea of restricting it to just RFC defined options.  We could validate them when passed in, and if we *really* wanted to, we could have a flag to disable the check (probably prefixed with `hack`).

The list of standard option tags is tracked by IANA: http://www.iana.org/assignments/sip-parameters/sip-parameters.xhtml#sip-parameters-4

In other news, I agree that `extraSupported` sounds good, because it matches `extraHeaders`.  Speaking of which, why not use a custom header for custom behavior? :-)

markandrus:
@wakamoleguy

> why not use a custom header for custom behavior?

Yes, this is an option, though the team I am working with wants to minimize our use of additional headers (specifically X- headers). After evaluating all of our options, and considering the spirit of the restriction RFC 3261 places on option tags:

> to prevent servers from insisting that clients implement non-standard, vendor-defined features in order to receive service

it seems like Supported is most appropriate for us and more semantic than, e.g., custom headers, Content-Types, or any other mechanism we could think of—especially since Supported option tags can be ignored, unlike Require.

I'd be happy to add whitelisting of option tags per @james-onsip's suggestion if y'all would like, but I'd _really_ like a `hack`, `nonStrict` mode, or some other escape hatch :-) I am not sure what such a flag should look like, though.

markandrus:
I've updated this PR with two commits: the first adds the IANA-registered SIP option tags to src/Constants.js and whitelists them in `getSupportedHeader`; the second adds a `hackAllowUnregisteredOptionTags` parameter to `UA` allowing you to bypass whitelisting.

PR #212, opened 2015-08-20
Emit "notify" from Session

markandrus:
I see that some incoming requests (BYE, REFER) and all outgoing requests emit a corresponding event on the `Session`. Would you consider adding incoming NOTIFY?

### DIFF
--- a/src/Session.js
+++ b/src/Session.js
@@ -723,6 +723,7 @@ Session.prototype = {
         break;
       case SIP.C.NOTIFY:
         request.reply(200, 'OK');
+        this.emit('notify', request);
         break;
     }
   },


joseph-onsip:
Can you provide some context for your use case? Generally, NOTIFYs are handled by the [`Subscription`](http://sipjs.com/api/0.7.0/subscription/#notify). Are you trying to handle a NOTIFY on the [implicit subscription](https://tools.ietf.org/html/rfc3515#section-2) created by a REFER?

markandrus:
Hi @joseph-onsip, I would like to handle NOTIFYs for subscriptions created by non-SUBSCRIBE mechanisms, as hinted at in [RFC 3265, Section 3.2](https://tools.ietf.org/html/rfc3265#section-3.2):

> Subscriptions are typically put in place using the SUBSCRIBE method; however, it is possible that other means have been used.
> ...
> If a NOTIFY request receives a 481 response, the notifier MUST remove the corresponding subscription even if such subscription was installed by non-SUBSCRIBE means (such as an administrative interface).

Catching the NOTIFYs in application-space seemed like a lightweight solution, though I could also imagine a mechanism that reuses SIP.js's `Subscription` mechanism without sending a SUBSCRIBE. What do you think?

wakamoleguy:
I'd be happy to accept this, but the notify should be replied to *before* emitting the event.  See `refer` and `bye` behavior above it.

markandrus:
Thanks @wakamoleguy—I have updated the PR

wakamoleguy:
Merged.  Thanks!

PR #218, opened 2015-09-09
WebRTC.MediaStreamManager - call MediaStreamTrack#stop instead of MediaStream#stop

joseph-onsip:
MediaStream.stop() is deprecated and will be removed in M47, around November 2015
    
see https://groups.google.com/d/msg/sip_js/t7Km0p11Jpk/rMmwJTCHAAAJ

### DIFF
--- a/src/WebRTC/MediaStreamManager.js
+++ b/src/WebRTC/MediaStreamManager.js
@@ -147,7 +147,9 @@ MediaStreamManager.prototype = Object.create(SIP.EventEmitter.prototype, {
     streams.forEach(function (stream) {
       var streamId = MediaStreamManager.streamId(stream);
       if (this.acquisitions[streamId] === false) {
-        stream.stop();
+        stream.getTracks().forEach(function (track) {
+          track.stop();
+        });
       }
       delete this.acquisitions[streamId];
     }, this);
--- a/test/helpers/HTML5.js
+++ b/test/helpers/HTML5.js
@@ -46,14 +46,21 @@
     }
   }
   getUserMedia.fakeStream = function () {
+    var audioTracks = [{
+      id: Math.random().toString(),
+      stop: jasmine.createSpy('stop'),
+    }];
+    var videoTracks = [];
     return {
-      getAudioTracks: function(id){
-        return [{
-          id: id
-        }];
-      }.bind(null, Math.random().toString()),
-      getVideoTracks: function(){return [];},
-      stop: jasmine.createSpy('stop')
+      getAudioTracks: function () {
+        return audioTracks;
+      },
+      getTracks: function () {
+        return audioTracks.concat(videoTracks);
+      },
+      getVideoTracks: function () {
+        return videoTracks;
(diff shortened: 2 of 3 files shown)


wakamoleguy:
That seems reasonable to me.  The one odd part is adding a spy in the helpers, but we do that in at least one other place.  As long as the spy methods are initialized from within the tests, rather than being statically set up beforehand, it should be fine.

joseph-onsip:
Yeah, that spy was already being added, just on the stream instead of the track: https://github.com/joseph-onsip/SIP.js/commit/269865bb6323e4d9f4f69c296e4da1de4eaa8f61#diff-de57ce79d9714a43d45b4fc11e14f7feL56

While we're on the subject of tests, though: I did try to have the `expect` calls happen within a `getTracks().forEach` loop, but the [second test](https://github.com/joseph-onsip/SIP.js/commit/269865bb6323e4d9f4f69c296e4da1de4eaa8f61#diff-abec1c99b2499f4e85bec100d7dfc2a2R176) was failing, so I just tested the first audio track (the only one  faked by `HTML5.js`). I did manually confirm in Chrome/Firefox that video tracks are stopped as well:

```javascript
var SIP = require('sip.js');
var ua = new SIP.UA({traceSip: true})
var session = ua.invite('welcome@onsip.com', document.createElement('video'));
session.on('accepted', setTimeout.bind(null, session.bye.bind(session), 5000));
```

wpp:
Hi,

FYI there are other changes coming up as well:

https://developers.google.com/web/updates/2015/07/mediastream-deprecations

- MediaStream.ended
- MediaStream.label
- MediaStream.stop()

are all deprecated in Chrome 45. And require the following changes:

- Use `MediaStream.active` to check if a MediaStream is streaming, not `MediaStream.ended`.
- Use `MediaStreamTrack.stop()` to stop streaming, not `MediaStream.stop().`
- If you need a unique identifier for a MediaStream use `MediaStream.id` instead of `MediaStream.label`. `MediaStreamTrack.label` provides a human-readable name for the source device for a stream, e.g. FaceTime HD Camera (Built-in) (05ac:8510).

joseph-onsip:
>  I did try to have the expect calls happen within a getTracks().forEach loop, but the second test was failing

I'm not sure what I was doing wrong earlier, but I got this working in https://github.com/joseph-onsip/SIP.js/commit/6a42ac5dd426689a840cf830fa3201c46ecf1562#diff-abec1c99b2499f4e85bec100d7dfc2a2

> FYI there are other changes coming up as well:

Thanks for the heads-up, @wpp. This PR addresses the `MediaStreamTrack.stop()` change. The `MediaStream.active` change is unnecessary, as there are no references to `MediaStream.ended` in the code. Likewise, the `MediaStream.id` change isn't necessary, but it might be useful here: https://github.com/onsip/SIP.js/blob/90a375873d277c53129da8283118fabac1c6906f/src/WebRTC/MediaStreamManager.js#L27

PR #224, opened 2015-09-15
Session#reject sends 486, not 480

joseph-onsip:
As mentioned at https://groups.google.com/d/msg/sip_js/Vp0xBn8MtUI/-sPuG969meIJ

### DIFF
--- a/src/ServerContext.js
+++ b/src/ServerContext.js
@@ -48,7 +48,7 @@ ServerContext.prototype.accept = function (options) {
 
 ServerContext.prototype.reject = function (options) {
   options = Object.create(options || Object.prototype);
-  options.statusCode || (options.statusCode = 480);
+  options.statusCode || (options.statusCode = 486);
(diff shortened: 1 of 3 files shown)


wakamoleguy:
Some links:

* [480](https://tools.ietf.org/html/rfc3261#section-21.4.18)
* [486](https://tools.ietf.org/html/rfc3261#section-21.4.24)
* [A "common scenario" for 486](https://tools.ietf.org/html/rfc3261#section-13.3.1.3)

486 as described explicitly states that the user is busy and is not willing or able to take additional calls.  The description for 480 also allows for things such as not wanting to be disturbed, application-level filtering, etc.  Unless we can say for certain that the user being busy (and not otherwise unavailable) is the default case when calling `session.reject()`, we definitely should not change this.

And...even if we do establish that, I don't believe it is worth it to change the default response code at this point, when the RFC and the API allow for both options.

PR #225, opened 2015-09-15
Add 2 unit tests for 423 response on REGISTER

ryan-rowland:
We're looking at implementing 423s in the future and wanted to make sure it's always handled correctly. I figured I'd commit them back to master as this functionality doesn't seem to be under harness yet.

### DIFF
--- a/test/spec/SpecRegisterContext.js
+++ b/test/spec/SpecRegisterContext.js
@@ -74,6 +74,34 @@ describe('RegisterContext', function() {
       expect(RegisterContext.send).toHaveBeenCalled();
     });
 
+    it('retries with the min-expires header on 423', function() {
+      RegisterContext.register(options);
+
+      var response = new SIP.IncomingResponse(ua);
+      response.status_code = 423;
+      response.cseq = RegisterContext.cseq;
+      response.setHeader('min-expires', 555555);
+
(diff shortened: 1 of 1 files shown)


wakamoleguy:
Thank you for the contribution!  It's always nice to see some new unit tests.  I have a couple sticking points, though.

My first nitpick would be that the mocked response objects seem a little gratuitous in their responses.  Are we really testing that we read the 'min-expires' header when requesting *any* header returns 123?  I'd much prefer to see a response object that is closer to the real thing and depends less on the specific implementation of `RegisterContext.receiveResponse`.

Secondly, I notice [the RFC says](https://tools.ietf.org/html/rfc3261#section-10.2.8) that the retry should contain an expires equal to *or greater than* the one specified in the 423 response.  I'd like the unit test to match that more closely.

ryan-rowland:
@wakamoleguy Fair enough. I've updated both tests to construct an IncomingResponse object and set the appropriate properties, and changed the test assertion to check for an `expires` equal to or greater than the `min-expires` header.

wakamoleguy:
Looks good, thanks!

wakamoleguy:
Whoops...sorry about the misclicks and spam updates.  It's probably time for me to go home.

Issue #166, opened 2015-04-08
Refactor hold/resume code to live in MediaHandler


olivermt:
Hi.

Still not quite there with making a pull request for our solution to reINVITE, but one thing we have discovered along the way is that the current handling of hold/resume is pretty messy.

I suggest moving all mangling/munging of the sdp to the mediahandler.

On setDescription it is very easy to store the SDP and inspect what changed if anything changed at all.

We can roll this into our new-mediahandler-per-transaction (if something changed that required an actual reINVITE or a handling of a reINVITE) way of doing things if the development team approves.

wakamoleguy:
Moving SDP mangling out of the Session sounds good.  One of the MediaHandler's main roles is to take the knowledge of SDP away from the Session.  Not only should the MediaHandler be able to use a different backend, which may not need mangling, but it should also be free to specify a completely different content type for session description.

I'm not sure how tightly couple this would be to your MediaHandler-per-transaction work.  This would certainly be a smaller change, overall, and could be added or merged more quickly than the larger branch.

markandrus:
@wakamoleguy I totally agree with

> Not only should the MediaHandler be able to use a different backend, which may not need mangling, but it should also be free to specify a completely different content type for session description.

I am currently looking at some "mixed/multipart" use cases described in [RFC 5261, Section 4.1](https://tools.ietf.org/html/rfc5621#section-4.1), and moving the message handling from `Session`—specifically the Content-Type header checks and `.body` accesses—into the `MediaHandler` would be a big win.

I think there may also be some interesting use cases for ORTC applications that do not natively support SDP. For example, once the `MediaHandler` takes over more of the description handling, an application could specify some JSON blob instead of SDP or use "multipart/alternative": 

> The transition from SDP to new session description protocols could be
> implemented using 'multipart/alternative' bodies.  SIP messages
> (e.g., INVITE requests) could carry a 'multipart/alternative' body
> with two body parts: a session description written in SDP and a
> session description written in a newer session description format.
> Legacy recipient UAs would use the session description written in
> SDP.  New recipient UAs would use the one written in the newer
> format.

Where is this on the SIP.js roadmap? Glancing over the code, it seems like step 1 could be to change the signature for `setDescription` to accept a `SIPMessage` and to add a method to get a description from a `SIPMessage`. The `getDescription` name is already taken, so maybe `getDescriptionFromMessage`?

egreenmachine:
[...]

Issue #160, opened 2015-03-06
Crusade against Utils.js


wakamoleguy:
I have strong, negative feelings towards the existence of **Utils.js**.  At best, it is a miscellaneous bag of functions with no strong connection between them.  At worst, it is a thorn in the side of dependency graphs everywhere.

In an ideal world, we could strip each function out of **Utils.js** and into either into smaller, semantically themed modules or re-home some functions to existing modules that use them.

This ticket is somewhat low priority, of course, as it should have no direct outward effect on the behavior of the library.

egreenmachine:
While  we would like to get rid of `Utils.js`, there is no problem with keeping it around. Due to the fact that it is not a bug, I am going to close this issue.

PR #231, opened 2015-09-22
Make Session SDP-agnostic

markandrus:
I have a use case where I need to be prepared to send and receive offers and answers in "multipart/mixed" messages. This use case may be too specific for SIP.js, so I looked at enabling it in a custom `MediaHandler`. Following up on #166, this PR includes changes that
- Move SDP hacks into `WebRTC/MediaHandler`.
- Expose an `onhold` property on `WebRTC/MediaHandler`—this is used in `Session` instead of matching a `RegExp` on the SDP.
- Remove SDP mangling from `sendReinvite`—now `WebRTC/MediaHandler` will check an `onhold` property in order to determine if it should rewrite the SDP in `getDescription`.
- Remove Content-Type checking from `Session`—now `Session` will pass the entire `SIPMessage` to `MediaHandler.hasDescription` (which can check its Content-Type, check if it has an empty `body`, or anything else).
- Pass the entire `SIPMessage` to `MediaHandler.setDescription`—for example, a `MediaHandler` expecting "multipart/mixed" messages needs to query the Content-Type in order to parse the `body`. There may be other `MediaHandler`s that observe other headers.
- Allow `MediaHandler.getDescription` to specify `extraHeaders` and `body`—for example, a `MediaHandler` could specify a Content-Type other than "application/sdp".

These are significant changes, and I only have the Jasmine tests passing (I have not yet tested with my application), but would you consider such a refactor?

### DIFF
--- a/src/Session.js
+++ b/src/Session.js
@@ -556,11 +556,13 @@ Session.prototype = {
 
     this.mediaHandler.setDescription(request)
     .then(this.mediaHandler.getDescription.bind(this.mediaHandler, this.mediaHint))
-    .then(function(body) {
-      request.reply(200, null, ['Contact: ' + self.contact], body,
+    .then(function(description) {
+      var extraHeaders = ['Contact: ' + self.contact];
+      extraHeaders = extraHeaders.concat(SIP.Utils.getDescriptionHeaders(description));
+      request.reply(200, null, extraHeaders, description.body,
         function() {
           self.status = C.STATUS_WAITING_FOR_ACK;
-          self.setInvite2xxTimer(request, body);
+          self.setInvite2xxTimer(request, description);
           self.setACKTimer();
 
           if (self.remote_hold && !self.mediaHandler.remote_hold) {
@@ -607,16 +609,16 @@ Session.prototype = {
 
     extraHeaders.push('Contact: ' + this.contact);
     extraHeaders.push('Allow: '+ SIP.UA.C.ALLOWED_METHODS.toString());
-    extraHeaders.push('Content-Type: application/sdp');
 
     this.receiveResponse = this.receiveReinviteResponse;
     //REVISIT
     this.mediaHandler.getDescription(self.mediaHint)
     .then(
-      function(body){
+      function(description){
+        extraHeaders = extraHeaders.concat(SIP.Utils.getDescriptionHeaders(description));
         self.dialog.sendRequest(self, SIP.C.INVITE, {
           extraHeaders: extraHeaders,
-          body: body
+          body: description.body
         });
       },
       function() {
@@ -768,7 +770,7 @@ Session.prototype = {
    * Response retransmissions cannot be accomplished by transaction layer
    *  since it is destroyed when receiving the first 2xx answer
    */
-  setInvite2xxTimer: function(request, body) {
+  setInvite2xxTimer: function(request, description) {
     var self = this,
         timeout = SIP.Timers.T1;
 
@@ -779,7 +781,10 @@ Session.prototype = {
 
       self.logger.log('no ACK received, attempting to retransmit OK');
 
-      request.reply(200, null, ['Contact: ' + self.contact], body);
+      var extraHeaders = ['Contact: ' + self.contact];
+      extraHeaders = extraHeaders.concat(SIP.Utils.getDescriptionHeaders(description));
+
+      request.reply(200, null, extraHeaders, description.body);
 
       timeout = Math.min(timeout * 2, SIP.Timers.T2);
 
@@ -1061,7 +1066,7 @@ InviteServerContext = function(ua, request) {
     .then(
       fireNewSession,
       function onFailure (e) {
-        self.logger.warn('invalid SDP');
+        self.logger.warn('invalid description');
         self.logger.warn(e);
         request.reply(488);
       }
@@ -1225,18 +1230,20 @@ InviteServerContext.prototype = {
       // Get the session description to add to preaccept with
       this.mediaHandler.getDescription(options.media)
       .then(
-        function onSuccess (body) {
+        function onSuccess (description) {
           if (this.isCanceled || this.status === C.STATUS_TERMINATED) {
             return;
           }
 
-          this.early_sdp = body;
+          this.early_sdp = description.body;
           this[this.hasOffer ? 'hasAnswer' : 'hasOffer'] = true;
+
+          extraHeaders = extraHeaders.concat(SIP.Utils.getDescriptionHeaders(description));
 
           // Retransmit until we get a response or we time out (see prackTimer below)
           var timeout = SIP.Timers.T1;
           this.timers.rel1xxTimer = SIP.Timers.setTimeout(function rel1xxRetransmission() {
-            this.request.reply(statusCode, null, extraHeaders, body);
+            this.request.reply(statusCode, null, extraHeaders, description.body);
             timeout *= 2;
             this.timers.rel1xxTimer = SIP.Timers.setTimeout(rel1xxRetransmission.bind(this), timeout);
           }.bind(this), timeout);
@@ -1254,7 +1261,7 @@ InviteServerContext.prototype = {
           }.bind(this), SIP.Timers.T1 * 64);
 
           // Send the initial response
-          response = this.request.reply(statusCode, reasonPhrase, extraHeaders, body);
+          response = this.request.reply(statusCode, reasonPhrase, extraHeaders, description.body);
           this.emit('progress', response, reasonPhrase);
         }.bind(this),
 
@@ -1301,14 +1308,14 @@ InviteServerContext.prototype = {
       iceServers,
       stunServers = options.stunServers || null,
       turnServers = options.turnServers || null,
-      sdpCreationSucceeded = function(body) {
+      descriptionCreationSucceeded = function(description) {
         var
           response,
           // run for reply success callback
           replySucceeded = function() {
             self.status = C.STATUS_WAITING_FOR_ACK;
 
-            self.setInvite2xxTimer(request, body);
+            self.setInvite2xxTimer(request, description);
             self.setACKTimer();
           },
 
@@ -1328,6 +1335,7 @@ InviteServerContext.prototype = {
 
         extraHeaders.push('Contact: ' + self.contact);
         extraHeaders.push('Allow: ' + SIP.UA.C.ALLOWED_METHODS.toString());
+        extraHeaders = extraHeaders.concat(SIP.Utils.getDescriptionHeaders(description));
 
         if(!self.hasOffer) {
           self.hasOffer = true;
@@ -1335,7 +1343,7 @@ InviteServerContext.prototype = {
           self.hasAnswer = true;
         }
         response = request.reply(200, null, extraHeaders,
-                      body,
+                      description.body,
                       replySucceeded,
                       replyFailed
                      );
@@ -1344,7 +1352,7 @@ InviteServerContext.prototype = {
         }
       },
 
-      sdpCreationFailed = function() {
+      descriptionCreationFailed = function() {
         if (self.status === C.STATUS_TERMINATED) {
           return;
         }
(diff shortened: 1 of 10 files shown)


src/Session.js
```diff
@@ -562,27 +548,27 @@ Session.prototype = {
...
+    this.mediaHandler.setDescription(request)
     .then(this.mediaHandler.getDescription.bind(this.mediaHandler, this.mediaHint))
-    .then(function(body) {
-      request.reply(200, null, ['Contact: ' + self.contact], body,
+    .then(function(description) {
+      var extraHeaders = ['Contact: ' + self.contact];
+      if (description.extraHeaders) {
+        extraHeaders = extraHeaders.concat(description.extraHeaders);
+      }
+      request.reply(200, null, extraHeaders, description.body,
```
vassjozsef: What is description.body?

markandrus: Yes, the idea is that `getDescription` returns enough information to be able to set a Content-Type header (and potentially other headers) using the `extraHeaders` property, and to set the body using the `body` property. I am not sure if this should be called `description`, but it is all I could think of at the time.

Another approach I thought of was to return a function that transforms the outgoing SIP message, but I don't think SIP.js is as well-suited to that approach.

markandrus:
Just following up: this is working pretty well in my application. I do not observe any behavior change. But one concern I have is that `hasDescription` might duplicate the work of `setDescription`. Maybe instead of

```js
if (hasDescription(request)) {
  setDescription(request);
}
```

Session.js should do something like

```js
var description = extractDescription(request);
if (description) {
  setDescription(description);
}
```

wakamoleguy:
Hey, I just wanted to post to say that this is definitely still on my radar.  I haven't had the time to really look at it closely yet.  And it's a big refactor, so I am keen to actually see it in action before merging it.  It looks like there is a lot of great work in here, and I want to make sure I can give it the attention it deserves.

markandrus:
Hi @wakamoleguy, thanks for the update.

>  I am keen to actually see it in action before merging it.

Yes, please take some time to review it first! :-) This is not a change I'd like to just push through, and I still have an open question about implementation (e.g. `hasDescription` vs `extractDescription`, as mentioned above). Normal INVITE and INVITE without SDP flows are working for me; however, I have not tested early media (I don't really have a use case for early media in my application).

james-onsip:
Overall, I think this is a great change. When we forked SIP.js, we did an initial mediaHandler refactor to attempt to remove sdp from Session, but the re-invite logic was added after that, so I'm glad to see it finally going where it belongs. I'm going to leave some notes about changes as I go:

- [1st commit](https://github.com/markandrus/SIP.js/commit/67737a42116684c3371781163f799177cde2916d) looks good

- [2nd commit](https://github.com/markandrus/SIP.js/commit/ba8f06866495283cf81eed720e28869191019f17): it looks like the mangle code will now get called in more situations. Specifically, whenever you are on hold and call getDescription (I believe an example would be in receiveReinvite). I don't believe this is problematic, but it is a processing change. I also commented on the diff, but I believe the `hold` variable in receiveReinvite may now be misused. Also, I'm not sure if `remote_hold` and `local_hold` were considered during these changes.

- [3rd commit](https://github.com/markandrus/SIP.js/commit/0d300e2e5519fe2dfe22477ea909dc560fffc17e): hasDescription is now more rigorous in many situations than it used to be. I think it's ok and we probably should have been checking both body and content-type, but there may be situations where that was not true.

- [4th commit](https://github.com/markandrus/SIP.js/commit/23dfee352b6569a061554b012ba151ed9ba5016a) looks good

- [5th commit](https://github.com/markandrus/SIP.js/commit/784cb36f3e5ee4004f11495552038b6937994c4d): I may need to brush up on the RFC here, but is Content-Type required? If the new `extraHeaders` property will always be at least Content-Type, and Content-Type is required, I'd much rather make it the property `contentType` and leave the `extraHeaders.push("Content-Type:"...)` lines. You can still pass `extraHeaders` back as a third property, but if Content-Type is always important, I would favor giving it it's own property. I think `description` is a good name for the object coming out.

In terms of `hasDescription` vs. `extractDescription`, it seems like you'd just be forcing a dependency between the two functions if you went with `extract` because of the object. If the idea of extracting the description was used elsewhere, I think it would be a good change, but if `extractDescription` will always be followed by `setDescription`, I'd rather `setDescription` be powerful enough to deal with the SIP message, and leave `hasDescription` as solely a validator. I could be missing the whole point here, so let me know if that makes sense.

Thanks again,
James

markandrus:
@james-onsip thank you for the review!

* Regarding the hold behavior:

    > I believe this edit actually changes the behavior incorrectly, unless it was moved elsewhere and I'm not seeing it. The hold variable was used to determine if the incoming request (a reinvite) wanted to put the call on or off hold. Now, I believe it is signaling whether or not the call has been placed on hold from our side. ...

    You may be right. I can't say I use the hold functionality, so I will have to revisit that portion of the code and get back to you.

* Regarding the 5th commit, that is a good suggestion. Per [RFC 3261, Section 20.15](https://tools.ietf.org/html/rfc3261#section-20.15),

    > The Content-Type header field MUST be present if the body is not empty.  If the body is empty, and a Content-Type header field is present, it indicates that the body of the specific type has zero length (for example, an empty audio file).

    I will update the return value. What do you think of

    ```js
    {
      // optional; can be string, null, or undefined
      body: null,
      // required if `body` is present and neither null nor undefined
      contentType: "application/sdp",
      // optional
      extraHeaders: [
        "foo: bar"
      ]
    }
    ```

* Thank you for your perspective regarding `hasDescription` vs. `extractDescription`. I agree it would create a dependency between the two functions in the return type of `extractDescription` and the type of the input to `setDescription`.

    > If the idea of extracting the description was used elsewhere, I think it would be a good change

    The only place I can think of using this functionality would be in unit tests, where you could create the input expected by `setDescription` without having to frame it in a SIP message. I don't feel strongly about it though.

    > if `extractDescription` will always be followed by `setDescription`, I'd rather `setDescription` be powerful enough to deal with the SIP message, and leave `hasDescription` as solely a validator. I could be missing the whole point here, so let me know if that makes sense.

    Letting `setDescription` be powerful enough to deal with the SIP message does make a simpler API, which I like. The only worry I have is that if `hasDescription` and `setDescription` duplicate some amount of work (for example, I have a custom MediaHandler that must parse a "multipart/mixed" message body) I may pay for it twice.

I'll see about working on these changes this week, getting things rebased, etc.

Thanks,
Mark

james-onsip:
I think using `contentType` as the property looks great.

In regards to the double work thing, since you are writing a custom MediaHandler, you could save the parsed message body on the MediaHandler object (`this`), then reuse it if it's there, or do it in another, non-interface function (should be possible, don't actually know though). I could imagine both `setDescription` and `hasDescription` call an internal `extractDescription` that does the parsing, but if an object has already been parsed then it is saved. at the end of `setDescription`, you can then clear it. It's not the best, but it does allow the parsing to only occur once. This may even be something to add to the interface, although for sdp I guess it would just kind of no-op.

markandrus:
@james-onsip thank you for you guidance on the hold functionality in the 2nd commit. I have rewritten it in such a way that I believe fixes the behavior (see commit message for more details). I've also updated the 5th commit based on our discussions and introduced a `getDescriptionHeaders` function to reduce code repetition.

Would you please take another look? Unit tests are passing, and I have successfully re-integrated these changes with my application, so I am fairly confident; however, I don't exercise the hold functionality in my own code. Is there a canonical test app I could test against?

james-onsip:
I like the changes for the second commit, although I think the usefulness of `remote_hold` and `local_hold` on the session object is at the point where they can probably be removed (albeit in another PR, they aren't entirely useless yet).

For the fifth commit, I think the utils function makes sense, but it certainly won't get past an author or two (*cough* @wakamoleguy and [the crusade](https://github.com/onsip/SIP.js/issues/160) *cough*). I'm not sure of exactly the best way, but based on your changes and quoting of the RFC, I would think making "Content-Type" not be an extraHeader would make sense. Making it default to `application/sdp`, but otherwise be set with an argument of some sort may be the best path.

What I describe seems like... a substantial change, that would most likely warrant more significant refactoring and necessitate a new release (and may not even make sense), so maybe just making the argument coming back the correct string ('Content-Type:application/sdp' and renaming the property to `contentTypeHeader` so as to obsolete the Utils function may be easiest to get this PR onto master.

I didn't think the idea(s) I listed above all the way through, so please let me know if I'm missing something or there is a better way (not involving Utils).

Also, a 'canonical test app' is a bit of a holy grail. We have a [demo phone](http://sipjs.com/demo-phone/) which is also on [github](https://github.com/onsip/sipjs.com) and can be built locally and altered, but even that is slightly out of date. I do not believe it currently has hold functionality, but that can probably be added fairly simply. If I have some time soon I can look into getting that set up for you.

Thanks,
James

markandrus:
> For the fifth commit, I think the utils function makes sense, but it certainly won't get past an author or two (cough @wakamoleguy and the crusade cough).

Ah, I see. I was wondering if it would make sense to instead define a local function in Session.js, but there might be a different way forward if I understand the rest of your comment correctly (more on that below).

> I'm not sure of exactly the best way, but based on your changes and quoting of the RFC, I would think making "Content-Type" not be an extraHeader would make sense.

Are you referring to the way that Content-Type is pushed onto `extraHeaders` before calling `sendRequest` or `reply`? This does feel a bit like a mismatch in the internal API.

> What I describe seems like... a substantial change, that would most likely warrant more significant refactoring and necessitate a new release (and may not even make sense)

Please let me know if I am off track here, but one thing I notice is that the object returned by `getDescription` now shares a couple properties (`body` and `extraHeaders`) with the `options` accepted by `sendRequest`. If we added support for the `contentType` property to `sendRequest`, then Session.js could simply merge the `options` it constructs with the object returned by `getDescription` before calling `sendRequest`. `reply` could be updated similarly.

Is this the sort of refactor you are envisioning?

> so maybe just making the argument coming back the correct string ('Content-Type:application/sdp' and renaming the property to contentTypeHeader so as to obsolete the Utils function may be easiest to get this PR onto master.

I am open to this, though it is not my preference. Interested to hear if `sendRequest` and `reply` should accept `contentType`… what do you think?

james-onsip:
I agree, that the smaller, "make the string the argument" was not as good as the larger refactor, but it may have been enough to get this by. I do prefer the larger refactor as well.

> Please let me know if I am off track here, but one thing I notice is that the object returned by getDescription now shares a couple properties (body and extraHeaders) with the options accepted by sendRequest. If we added support for the contentType property to sendRequest, then Session.js could simply merge the options it constructs with the object returned by getDescription before calling sendRequest. reply could be updated similarly.

This would definitely be the cleanest, preferred way to do it, and hopefully the conclusion I would've reached had i thought about it more. Giving contentType it's own property in the object passed to `sendRequest` and `reply` certainly feels more in line with how the RFC describes it. I think passing it through and merging with options is good, certainly better than what is there now, but in the future maybe more could be done.

(This paragraph may be overkill and unnecessarily complicated, I just wanted to put it down as potential for the future).
I'm thinking it may make sense for it to work like [route_set](https://github.com/onsip/SIP.js/blob/master/src/SIPMessage.js#L87-L91). If you are using a custom MediaHandler that can only handle a specific type of message, making a config option that defines what type is acceptable may make sense. Maybe just making it a property of the MediaHandler that could be read when necessary? And, just like route set, this could be overridden by the options merging idea you describe above. This way, the contentType argument becomes optional, and contentType is coded to be  more like an RFC MUST. If that doesn't make any sense, doing it as you described would still work just great.

Thanks,
James

markandrus:
> This would definitely be the cleanest, preferred way to do it, and hopefully the conclusion I would've reached had i thought about it more. Giving contentType it's own property in the object passed to sendRequest and reply certainly feels more in line with how the RFC describes it. I think passing it through and merging with options is good, certainly better than what is there now, but in the future maybe more could be done.

I will give this a shot. It may be a little bit before I have the next iteration of this PR.

Regarding your route_set idea, I am not sure I understand fully, so I think I will punt on for now :-)

Thanks,
Mark

egreenmachine:
I did some work to address the remaining outstanding issue that was preventing this from being merged as well as updating it to the latest version of SIP.js. It is on my local branch. If it looks good, I can go ahead and merge it and then close this PR. https://github.com/egreenmachine/SIP.js/tree/mediahandler-refactor-egreen

markandrus:
@egreenmachine that's awesome! I ended up pretty busy with other tasks and this implementation was "good enough" for my fork, so I never got around to rewriting this. I'll take a look at your branch later today and compare.

egreenmachine:
I ran through some tests on several apps we have here and it all looked good. Even hold/unhold renegotiation worked. @markandrus did you get a chance to drop it into your application?

markandrus:
@egreenmachine yes, I pulled your branch, reviewed your changes, applied my patches, and re-tested with my application—it works correctly! Please close this PR and merge your updates :-)

Issue #232, opened 2015-09-28
expire not taken into account in 200 of SUBSCRIBE


matthieusieben:
[RFC6665](http://tools.ietf.org/html/rfc6665#section-3.1.1) describes that 200 responses to SUBSCRIBE must contain a `Expires` header.

It looks like SIP.js ignores this value and only checks for an expire timeout in the NOTIFY's `Subscription-State` header. If the notify does not contain an expire, then the subscription is never refreshed.

`setExpiresTimeout` (in [Subscription.js](https://github.com/onsip/SIP.js/blob/master/src/Subscription.js#L200)) should use the `Expires` header from the SUBCRIBE's 200 if no expire is present in the NOTIFY's `Subscription-State` header.

matthieusieben:
The section about the [Notifier NOTIFY Behavior](https://tools.ietf.org/html/rfc3265#section-3.2.2) only states that NOTIFYs **SHOULD** (and not **MUST**) contain an "expires" parameter. So using the value from the SUBSCRIBE's 200-class response seems more reliable.

matthieusieben:
Actually, it looks like the `Expires` header is taken into account to set a refresh timeout ([here](https://github.com/onsip/SIP.js/blob/master/src/Subscription.js#L99)). But as soon as a NOTIFY is received, the timeout is [cleared](https://github.com/onsip/SIP.js/blob/master/src/Subscription.js#L219) and not re-created if the NOTIFY did not contain a `expire` param in the `Subscription-State`.

wakamoleguy:
Another good catch!  The PR request (#236) looks good.  I'll accept that shortly.

PR #233, opened 2015-09-28
Resolves onsip/SIP.js#232

matthieusieben:
Refresh subscriptions even if the NOTIFY requests do not contain `expires` param in the `Subscription-State` header.

### DIFF
--- a/src/Subscription.js
+++ b/src/Subscription.js
@@ -240,6 +240,8 @@ SIP.Subscription.prototype = {
           switch (sub_state.reason) {
             case 'deactivated':
             case 'timeout':
+              this.terminateDialog();
+              this.request = new SIP.OutgoingRequest(this.method, this.request.to.uri.toString(), this.ua, null, this.extraHeaders.slice());
               this.subscribe();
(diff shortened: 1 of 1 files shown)


james-onsip:
Hi matthieusieben,
  Thanks for submitting this PR. I need to dig back into the code a little more before I'm sure if what you've changed is what is definitely needed, but I can at least point out a few initial issues.

It looks like the travis build has failed, because some of the existing tests failed. I didn't look too closely, but it looked like the tests that failed are looking for general subscription behavior and should pass if subscriptions are functioning in a proper, RFC-compliant manner. If these tests are in fact written to test incorrect things, please update them with an explanation as to what was incorrect about them.

In terms of what broke the tests, it might be the new variable you are defining (expires), is not set in the 'if' portion of the conditional, only the 'else'. The timeout is then set based on this variable, so in the 'if' case we will be using an undefined variable for the timeout.

Good luck,
James

matthieusieben:
I believe #236 covers this issue in a more adequate way.

Issue #234, opened 2015-09-29
Same dialog is reused when receiving Subscription-State is terminated;reason=timeout


adaxi:
Here is the NOTIFY informing that our subscription is terminated:

```
NOTIFY sip:cbsr5umj@ss5fdi3s9p32.invalid;transport=ws SIP/2.0
Via: SIP/2.0/WS 172.16.10.33:5060;branch=z9hG4bK2b37d1c6
Max-Forwards: 70
From: <sip:ID00000002@uep>;tag=as114a0625
To: <sip:6d2b467e-3234-4a6f-81a1-af0edab67638@escaux.com>;tag=6vd6h5hbf8
Contact: <sip:ID00000002@172.16.10.33:5060;transport=WS>
Call-ID: 9qaqhhdfavpjlj0np4fj
CSeq: 112 NOTIFY
User-Agent: ESCAUX-SIP
Subscription-State: terminated;reason=timeout
Event: presence
Content-Type: application/epresence+json
Content-Length: 0
```

Here is the SUBSCRIBE  sent back by SIP.js:

```
SUBSCRIBE sip:ID00000002@uep SIP/2.0
Via: SIP/2.0/WSS ss5fdi3s9p32.invalid;branch=z9hG4bK586217
Max-Forwards: 70
To: <sip:ID00000002@uep>
From: <sip:6d2b467e-3234-4a6f-81a1-af0edab67638@escaux.com>;tag=6vd6h5hbf8
Call-ID: 9qaqhhdfavpjlj0np4fj
CSeq: 2530 SUBSCRIBE
Authorization: Digest algorithm=MD5, username="6d2b467e-3234-4a6f-81a1-af0edab67638", realm="asterisk", nonce="33c25cbe", uri="sip:ID00000002@uep", response="36f2a49c2e05a24c048723e7fdc75033"
Accept: application/epresence+json
Event: presence
Expires: 60
Contact: <sip:cbsr5umj@ss5fdi3s9p32.invalid;transport=ws>
Allow: ACK,CANCEL,BYE,OPTIONS,INFO,NOTIFY,INVITE,MESSAGE
Supported: outbound
User-Agent: Escaux Connect - Development
Content-Length: 0
```

As you can see the Call-ID and from tag are identical. 
The expected outcome would be the creation of a new dialog.

wakamoleguy:
Good catch on this one.  For those looking on, here are some relevant [RFC 6665](https://tools.ietf.org/html/rfc6665) snippets:

[...]

PR #235, opened 2015-09-29
Resolves onsip/SIP.js#234

adaxi:
The first line ensures that the dialog is terminated.
The second line prepares the new request, which will have a new Call-ID and new from tag.

### DIFF
--- a/src/Subscription.js
+++ b/src/Subscription.js
@@ -237,6 +237,8 @@ SIP.Subscription.prototype = {
           switch (sub_state.reason) {
             case 'deactivated':
             case 'timeout':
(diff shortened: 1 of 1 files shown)


wakamoleguy:
See #234 for comments on the issue.  I'd like to proceed carefully with this one, as a lot of this code was not planned to be run when changing underlying parameters.  Specifically, by changing `this.request`, we are counting on the dialog to change as a side-effect, rather than explicitly terminating the subscription and creating a new one.  That feels dangerous, and it's very hard to test reliably.

I'm going to close this PR for now, until we decide what form the solution should take, if any.

Issue #238, opened 2015-10-14
turnServer.urls must be an array, not a string


wackazong:
According to http://sipjs.com/api/0.5.0/ua_configuration_parameters/#turnservers, turnServers.urls can be a string or an array. 

But for the code in https://github.com/onsip/SIP.js/blob/7e7df079b8f914dfff259ea0c03830bea565033e/src/WebRTC/MediaHandler.js#L56 it is required to be an array.

I think the documentation needs to be updated here.

egreenmachine:
Hmmm. This looks like it is actually a bug with SIP.js introduced: https://github.com/onsip/SIP.js/commit/8c0a121bc47988c626662dcb70659c1e282a9e60. This commit looks like it was to fix a Firefox issue with how we are passing TURN servers to the browser. We should determine if we want to still support strings in the turn url or not. I think that we should support strings and we could resolve this issue by then forcing strings into an array.

wackazong:
https://github.com/onsip/SIP.js/pull/239

wakamoleguy:
Have we tested if Firefox now accepts arrays?

wakamoleguy:
Also, turn servers are put through [configuration parsing](https://github.com/onsip/SIP.js/blob/master/src/UA.js#L1504), which looks like it will already force single URLs into array of URLs.

Have you actually seen this happening?  What are the symptoms?

wackazong:
You are right, we were modifying the turnservers directly. [Configuration parsing] (https://github.com/onsip/SIP.js/blob/master/src/UA.js#L1504) already handles this.

PR #239, opened 2015-10-14
Force turnServers.urls into an array

wackazong:
If a single server is given a string can be used which must here be forced into an array.


wackazong:
Fixes #238

wakamoleguy:
Rather than use `toString` to test if it is not an Array, I'd much rather enforce the 'string or Array' logic using typeof and instance of, respectively, as we do elsewhere in the library (see UA configuration parsing, for example).

That being said, I posted some questions on Issue #238 itself that should be answered before we consider a pull request.

Issue #249, opened 2015-11-09
Enable video after start only audio call


Slawutich:
I'm sorry can not find in the documentation
how to request a new SDP with the included video and send re-invite, without completing the current only audio call
I start a call like this

``` JavaScript
var options = {
        media: {
            constraints: {
                audio: {
                    optional:[
                        {googEchoCancellation:true},
                        {googAutoGainControl:true},
                        {googNoiseSuppression:true},
                        {googHighpassFilter:true},
                        {googAudioMirroring:false},
                        {googNoiseSuppression2:true},
                        {googEchoCancellation2:true},
                        {googAutoGainControl2:true},
                        {googDucking:false}
                    ]
                },
                video: false
            },
            render: {
                remote: this.getRemoteVideo(),
                local: this.getLocalVideo()
            }
        }
    };

var session = ua.invite('sip:'+number+'@'+host, options);
```

markandrus:
@Slawutich you can add a MediaStream to the PeerConnection and then call `session.sendReinvite`. You can get at the PeerConnection with `session.mediaHandler.peerConnection`.

Slawutich:
Thank you very much
but it seems that for the new video stream are no candidates in the SDP
I'll look again, I can somewhere made a mistake

rscreene:
I think I'm facing the same problem as you:  the re-INVITE is being sent before the video candidates are collected.
Before I go digging deep into the code, did you manage to find the problem?

Slawutich:
no
I was very busy, so it was not possible to solve the problem

rscreene:
Thanks @Slawutich I'll try to get time to delve into the code.

Slawutich:
onIceCompleted does not wait for the selection of candidates and immediately called
```JavaScript
var mediaHandler = session.mediaHandler;
 
mediaHandler.mediaStreamManager.acquire({media: {
	constraints: {
		audio: false,
		video: true
	}
}})
.then(function acquireSucceeded(streams) {
	/**
	 * add to local media, after the end of call stream must be deleted
	 */
	mediaHandler.mediaStreamManager.release(mediaHandler.localMedia);
	mediaHandler.peerConnection.removeStream(mediaHandler.localMedia[0]);
	mediaHandler.localMedia = streams;

	mediaHandler.addStreams(streams)
	.then(function streamAdditionSucceeded(streams){
		/**
		 * reset onIceCompleted, after setLocalDescription he immediately called
		 * without waiting for the collection of candidates
[...]

PR #259, opened 2015-12-10
Enable video after start only audio call

rscreene:
My attempt at fixing #249 - the re-INVITE is being sent before the video candidates are collected.

I had to do make a couple of changes:
1) Added a media option to the sendReinvite() method so to add video to an audio call you just have to do: `session.sendReinvite({media: { constraints: { video: true } })`
2) Initialised the onIceCompleted promise whenever ICE is started.
3) In MediaHandler.getDescription() set the iceRestart attribute in the RTC constraints if it is necessary.  (I have overloaded the attribute that may be passed to the createOffer() method of PeerConnetion for this).

It works for our requirements but I cannot guarantee that I haven't broken anything.

Please do what you wish to this PR.

### DIFF
--- a/src/Session.js
+++ b/src/Session.js
@@ -606,6 +606,12 @@ Session.prototype = {
 
   sendReinvite: function(options) {
     options = options || {};
+    options = Object.create(Session.desugar(options));
+    SIP.Utils.optionsOverride(options, 'media', 'mediaConstraints', true, this.logger, this.ua.configuration.media);
+    
+    if (options.media) {
+      this.mediaHint = options.media;
+    }
 
     var
       self = this,
--- a/src/Session.js
+++ b/src/Session.js
@@ -608,7 +608,7 @@ Session.prototype = {
     options = options || {};
     options = Object.create(Session.desugar(options));
     SIP.Utils.optionsOverride(options, 'media', 'mediaConstraints', true, this.logger, this.ua.configuration.media);
-    
+
     if (options.media) {
       this.mediaHint = options.media;
     }
--- a/src/UA.js
+++ b/src/UA.js
@@ -683,14 +683,6 @@ UA.prototype.receiveRequest = function(request) {
          * and without To tag.
          */
         break;
-      case SIP.C.NOTIFY:
-        if (this.configuration.allowLegacyNotifications && this.listeners('notify').length > 0) {
-          request.reply(200, null);
-          self.emit('notify', {request: request});
-        } else {
-          request.reply(481, 'Subscription does not exist');
-        }
-        break;
       default:
         request.reply(405);
         break;
@@ -922,9 +914,7 @@ UA.prototype.loadConfig = function(configuration) {
 
       authenticationFactory: checkAuthenticationFactory(function authenticationFactory (ua) {
         return new SIP.DigestAuthentication(ua);
-      }),
-
-      allowLegacyNotifications: false
+      })
     };
 
   // Pre-Configuration
@@ -1161,7 +1151,6 @@ UA.configuration_skeleton = (function() {
       "media",
       "mediaConstraints",
       "authenticationFactory",
-      "allowLegacyNotifications",
 
       // Post-configuration generated parameters
       "via_core_value",
@@ -1601,13 +1590,7 @@ UA.configuration_check = {
       }
     },
 
-    authenticationFactory: checkAuthenticationFactory,
-
-    allowLegacyNotifications: function(allowLegacyNotifications) {
-      if (typeof allowLegacyNotifications === 'boolean') {
-        return allowLegacyNotifications;
-      }
-    }
+    authenticationFactory: checkAuthenticationFactory
   }
 };
 
--- a/src/WebRTC/MediaHandler.js
+++ b/src/WebRTC/MediaHandler.js
@@ -391,6 +391,12 @@ MediaHandler.prototype = Object.create(SIP.MediaHandler.prototype, {
       self._remoteStreams.push(e.stream);
       self.render();
       self.emit('addStream', e);
+
+      e.stream.onaddtrack = function (te) {
+        self.logger.log('track added: '+ te.track.id);
+        self.render();
+        self.emit('addTrack', te);
+      };
     };
 
     this.peerConnection.onremovestream = function(e) {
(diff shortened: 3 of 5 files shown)


rscreene:
Mmmm, this works fine for Firefox Nightly.  I'm having problems with Chrome.  The stream is set up (I can see it in the statistics), but the onaddstream does not fire.

rscreene:
I have been unable to get this working on Chrome.  So closing this PR.

rscreene:
Got it working on Chrome by handling the stream onaddtrack event.  Re-opening PR.

james-onsip:
- [1st commit](https://github.com/rscreene/SIP.js/commit/1e345a568e0f27a02993dd7578feb1684004d348) looks good

- [2nd commit](https://github.com/rscreene/SIP.js/commit/d71e13e95e1072d1b7c933789f1c74e210789028): I see that iceRestarting was taken off of the RTCConstraints in a later commit, that is definitely preferred to how this commit approached it. Also, it seems that ice will restart if ANY of the conditions are not met, not just the iceRestart (or moreCandidates) variable. Was this intended?

- [3rd commit](https://github.com/rscreene/SIP.js/commit/64b2794d0a14558bc8519802566af718bcac8bd9) looks good, it just corrects an issue I had with the 2nd commit.

- [4th commit] (https://github.com/rscreene/SIP.js/commit/4211228a0f06e44a1b6c14ee6f06a69cd7831b80): While I do think this is a preferable change (as we've dealt with it in our own apps), this may lead to some unexpected behavior. I don't think it would break anything, and if anything is probably expected by users, but holding off on this until 0.8.0 may be necessary.


The PR looks great, thanks for putting the time in to making this work. After clearing up the question I have, as well as thinking about the `onaddtrack` a bit more, I'd say it is ready to be merged. Also, I made some changes to ICE Handling yesterday that may be useful to you, although it would probably just be overkill.

Thanks,
James

rscreene:
@james-onsip Thanks for your comments.  With regards to your question I think the behaviour is fine.  The comment is a bit mis-leading it does not actually invoke ICE restart - ICE restart will be performed automatically  by pc.setLocalDescription().  The else clause simply set the promise when the ICE completes.  There is a corner case where a re-negotiation occurs: the promise will be re-set, but I can't see this having any problems.  Let me know if this doesn't make sense!

james-onsip:
@rscreene,
  Now that I look more at the code, you're right, but there is a difference in functionality, albeit slight. I believe your change is correct for what you are attempting, but either way, I think dropping that whole bit of code is safer.

  The difference in functionality occurs in that now, an ICE collection MUST be occurring in order for `initIceCompleted`'s promise to resolve, whereas before, as long as it completed at any point prior, the code could continue. This is due to the new `deferred` being created for `onIceCompleted`. I guess this used to make sense as you were only worried about onIceCompleted happening once, but with your change it gets a little dicey setting the new `deferred` up AFTER the processing begins. The conditional should guard against this, but I think it can be avoided all together.

What if you were to call `initIceCompleted` before the call to `pc.setLocalDescription` (line 481 or so)? That way, the entire `then` function could just be replaced by `self.onIceCompleted.promise` (I think that should work anyway. That feels safer and cleaner, but I may be missing something. What do you think?

rscreene:
@james-onsip I think you might be right!  If you would like to make the changes I should be able to test them in our environment next week.

james-onsip:
[patchfile.txt](https://github.com/onsip/SIP.js/files/99299/patchfile.txt)

There's the diff/patch that should make the change we discussed. Maybe there's' a better way to do this, but  I couldn't figure it out through github so I figured this works fine. Also, you may want to catch the branch up to master before applying this and retesting, as there is a conflict. A lot of the MediaHandler initialization stuff was moved into a function so it can be reused, and the ICE stuff was also reused, so maybe your code will actually clean up some of that. Let me know if you need a hand with that conflict.

rscreene:
I sitll haven't got my test environment set-up as yet.  I'll try later in the week.

rscreene:
I've tried to get this patch working but failed! (When created an audio-only call I get a PeerConnection failure).  I'll try to debug when I get some free time.

james-onsip:
[patchfile.txt](https://github.com/onsip/SIP.js/files/112388/patchfile.txt)

I think the merge conflicts were resolved incorrectly, and it got rid of one of the functions you made (`initIceCompleted`). I made a new patch, and was able to get an audio call working with it. 

Thanks,
James

soulhunter1987:
Hi. When do you expect to include it into a stable release?

james-criscuolo:
I see we haven't updated this in a while, but that's given it ample time to be tested. Any luck with that, @rscreene?

rscreene:
I'm afraid I got dragged onto something else before I managed to merge. TBH I don't think i'm going to have time to finish it in the immediate future.   Sorry about that.

rscreene:
It looks like I'm able to look at it again.  I'll keep you posted...

rscreene:
Sorry - closed in error!

rscreene:
I've managed to get something working for this now.

The patch supplied by @james-criscuolo works well.  However, I have had to skip a single unit test since the promise always timed-out.

I have used only the PC.onaddstream and Stream.onaddtrack events since these are all that are required for my application (calling FreeSWITCH from Firefox and Chrome).  It will be worth implementing the newer events identified by @soulhunter1987 in #249

soulhunter1987:
What about merging this pull request? If you say it is working... PC.addstream is deprecated and will be removed, but not wery soon till all current WebRTC applications are based on it.

james-criscuolo:
There are two things preventing this from getting merged:

1) Why does that unit test fail? Does that mean that functionality no longer works?

2) The commits can be cleaned up a bit. Currently there are 10, looks like there can be significantly less.

wpp:
Hi all,

FYI: I checked out this branch (rscreene:master) and tested it with a recent Chrome version (51.0.2704.106).

After starting a audio only call **I could not** get the re-invite to work via:

    session.sendReinvite({{media: { constraints: { video: true } });

The initiating side got "stuck" at `this.mediaHandler.getDescription(self.mediaHint)` and never sent the Invite request.

soulhunter1987:
@rscreene ?

rscreene:
My apologies, I haven't had time to get back onto this.  It doesn't look like its going to happen in the short term either.  Sorry for not letting you know sooner.

egreenmachine:
This feature is now covered in [0.8.0](https://github.com/onsip/SIP.js/releases/tag/0.8.0). I wrote a [renegotiation guide](https://sipjs.com/guides/renegotiation/) to show users how to do this. 

While we ended up going a different route, we appreciate the effort you put into SIP.js and we hope that you continue to use and contribute to the project. Thanks

Issue #258, opened 2015-12-08
Call Dropping after 60 Seconds on 0.7.2 (Upgraded from 0.7.0)


lylepratt:
We have calls dropping after 60 seconds after upgrading to 0.7.2 from 0.7.0. It seems to be related to session timers. Can you help us understand why this is happening now and any potential solutions?

thomasquintana:
+1

egreenmachine:
Can you provide [Gists](https://gist.github.com/) of your SIP.js logs and FreeSWITCH logs?

lylepratt:
Freeswitch is returning a "500" when the call dies from the Sip.js side. I'll try to get some logs, but we'll need to set up a test environment for this first.

lylepratt:
Here is the GIST from the Sip.js side:
https://gist.github.com/lylepratt/50e9b645b7ab47c58270

One thing that was RED in javascript console after turning log level to debug: 
`Tue Dec 08 2015 12:34:51 GMT-0600 (CST) | sip.invitecontext.mediahandler | 6hili6vcbbfsf3cl8113scasefacqp | Failed to set local answer sdp: Failed to push down transport description: Offerer must use actpass value for setup attribute.`

josephfrazier:
After browsing through https://github.com/onsip/SIP.js/compare/0.7.0...0.7.2, I'm guessing this might have something to do with https://github.com/onsip/SIP.js/commit/bd5c3b21db50c33036c9b229b1581ea7a327d45e (proposed in #170 to fix #168)

lylepratt:
Yes it appears that that is the change which is causing this.

josephfrazier:
I dug into this one for a while, and I think I've gotten down to the root causes of it. Here's what I've come up with:

Before bd5c3b21db50c33036c9b229b1581ea7a327d45e, sip.js would automatically reply with a 488, which freeswitch interpreted as allowing the session to continue, in accordance with [page 16 of RFC 3261](https://tools.ietf.org/html/rfc3261#page-16):

>    During the session, either Alice or Bob may decide to change the
   characteristics of the media session.  This is accomplished by
   sending a re-INVITE containing a new media description.  This re-
   INVITE references the existing dialog so that the other party knows
   that it is to modify an existing session instead of establishing a
   new session.  The other party sends a 200 (OK) to accept the change.
   The requestor responds to the 200 (OK) with an ACK.  If the other
   party does not accept the change, he sends an error response such as
[...]

PR #262, opened 2015-12-14
Add hackReinviteErrorCode UA option

josephfrazier:
This allows issues like https://github.com/onsip/SIP.js/issues/258 to be
worked around by adding the following line to the UA configuration:

      hackReinviteErrorCode: 488,

(@lylepratt, could you confirm that this works for you?)

### DIFF
--- a/src/Session.js
+++ b/src/Session.js
@@ -593,7 +593,11 @@ Session.prototype = {
     })
     .catch(function onFailure (e) {
       var statusCode;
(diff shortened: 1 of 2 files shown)


egreenmachine:
I do not think we should merge this as it was fixed in a newer version of FreeSWITCH https://github.com/onsip/SIP.js/issues/258#issuecomment-164578265

josephfrazier:
Agreed. It's much better to have this fixed server-side, where the problem was.

Issue #272, opened 2016-01-14
Local media streams should not be required


watsoncj:
Since the PeerConnection does not require a local stream for receive-only use case, I'd like a way to invite without having to create a local stream.

Would you be open to a pull-request to support media constraints of `video: false, audio: false`?

The current behavior when these constraints are passed is an error from gUM:
````
DOMException: Failed to execute 'webkitGetUserMedia' on 'Navigator': At least one of audio and video must be requested
````

The proposed change is: 

When `video: false, audio: false` is explicitly provided:

1.  Do not attempt to acquire local media.
2. Do not add any streams to peerConnection.

wakamoleguy:
I'd be open to a pull request for that, sure.  Ideally, it should play nicely with the OfferToReceiveAudio/Video constraints.  Right now, we assume that you only want to receive things for which you requested access.  So I would expect `{audio: false, video: false}` to require those constraints to be specified.

watsoncj:
Thanks, good point!

Are you alright with the offerToReceiveAudio/Video constraints still needing to be put on the `mediaHandler.RTCConstraints` manually? 

This would be independent of the `invite.option.media.constraints`. Primarily because SIP.js doesn't have any knowledge and just passes them to `peerConnection.createOffer`.

wakamoleguy:
I think that makes sense, yeah.

watsoncj:
Thanks for the guidance. Don't hesitate to let me know if the PR needs any work.

PR #273, opened 2016-01-14
Do not require local streams.

watsoncj:
Use following media constraints to avoid getUserMedia call:

{video: false, audio: false}

Should be used in conjunction with mediaHandler.RTCConstraints of offerToReceiveAudio/Video.

Fixes https://github.com/onsip/SIP.js/issues/272


james-onsip:
Looks great, thanks!

PR #280, opened 2016-01-25
Run Travis against Node LTS and Stable versions

josephfrazier:
See commit messages for references.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ language: node_js
 node_js:
   - "0.10"
   - "4.2"
+  - "stable"
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ language: node_js
 
 node_js:
   - "0.10"
+  - "4.2"
 
 cache:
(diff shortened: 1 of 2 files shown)


markandrus:
+1 for the grunt-contrib-jasmine bump—just ran into this testing with Node 5/NPM 3

james-onsip:
While doing some updates for 0.7.3, I noticed that our websockets library (github.com/websockets/ws) has gone through several iterations since we've updated, and they've dropped support for node 0.8. I believe that is currently the lowest number we support, and I would be in favor of dropping that to make way for the ws version bump, which is our only non-dev, non-optional dependency listed.

This would probably require a bit of testing before getting on master, but I think it'd be for the better.

josephfrazier:
Oh, do we even advertise support for Node 0.8? I don't think I've ever tried it, haha. Anyway, we should definitely consider upgrading `ws` (and our other dependencies). Want to open a pull request for it? I'm going to merge/close this one.

james-onsip:
sure, I'll make a PR and get the process started of testing

PR #282, opened 2016-01-26
Update ws dependency to 1.0.1

james-criscuolo:
Note: this drops support for node < 0.12, so travis and package.json adjusted as well

This is a pretty big change, as several versions of node support are dropped, but keeping up with the ws dependency is most likely good in the long run.

This is yet to be tested, but testing will begin today.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ sudo: false
 language: node_js
 
 node_js:
-  - "0.10"
+  - "0.12"
   - "4.2"
   - "stable"
 
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "pegjs": "^0.8.0"
   },
   "engines": {
-    "node": ">=0.8"
+    "node": ">=0.12"
   },
   "license": "MIT",
   "scripts": {
@@ -50,7 +50,7 @@
     "test": "grunt travis --verbose"
   },
   "dependencies": {
-    "ws": "^0.6.4"
+    "ws": "^1.0.1"
   },
   "optionalDependencies": {
     "promiscuous": "^0.6.0"


wakamoleguy:
Concept ACK.  In general, we should try to move support to at least the LTS version of Node.js.  Dropping support for versions before that is minimal pain.

josephfrazier:
:+1:

Here are the [`ws` release notes](https://github.com/websockets/ws/releases). A couple of things I noticed:

> * Removed the client code. It was simple wrapper that really doesn't belong in a full ledged node.js library. If you want browser support you could just conditionally import it the WebSocket server. `var WS = window.WebSocket || require('ws')`

This doesn't affect the `dist/` version of SIP.js, nor should it affect Node users that simply `require('sip.js')` (see https://github.com/onsip/SIP.js/pull/108 for more info on how this works). However, the following (very edge, I'd think) case does not work _when browserified_:

```
[josephfrazier@Josephs-MacBook-Pro ~/workspace/sip.js] (james-onsip-master 1U) $
echo; cat test.js; echo

// get default environment
var environment = require('./src/environment');

// override with custom implementations
environment.WebSocket = require('ws');

// use custom environment with SIP.js
var SIP = require('./src/SIP')(environment);
var ua = new SIP.UA({traceSip: true});

[josephfrazier@Josephs-MacBook-Pro ~/workspace/sip.js] (package.json-ws-browser-field 1U) $
browserify test.js | smokestack
Error: Cannot find module 'bufferutil' from '/Users/josephfrazier/workspace/web/web-webrtc-phone/frameworks/sip.js/node_modules/ws/lib'
    at /usr/local/lib/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:50:17
    at process (/usr/local/lib/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:119:43)
    at /usr/local/lib/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:128:21
    at load (/usr/local/lib/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:60:43)
    at /usr/local/lib/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:66:22
    at /usr/local/lib/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:21:47
    at FSReqWrap.oncomplete (fs.js:82:15)
```

We could add some more code to keep this working (see https://github.com/websockets/ws/commit/e54d45fbab3b19c2940e9057ce1e7b8f105873e0 for conversation about the breaking change in `ws`, and [here](https://github.com/defunctzombie/package-browser-field-spec#replace-specific-files---advanced) for the general form that a workaround could take), but I'm not sure if it's worth it.

> * Binary addons have been completely removed. Even as optional dependency, it seems that npm is just unable to properly handle the builds causing installations to fail and lead to massive developer issues. While the dependencies have been removed, you can still optionally install things. See https://github.com/websockets/ws#opt-in-for-performance for additional information.

I don't know that any sip.js users have experienced build issues like this, so we may want to consider adding `bufferutil` and/or `utf-8-validate` into package.json. On the other hand, I'm thinking it's better to leave this decision up to the sip.js user.

james-criscuolo:
Testing for this was successful, and we are now able to accomodate this PR, so merging.

Issue #290, opened 2016-02-20
Session's Hold/Unhold methods should allow to pass options to sendReinvite


kirill-konshin:
Currently it's impossible to understand whether hold/unhold was successful because `hold` event is fired before any actual communication with the server.

Ideally both methods should return a Promise.

PR #291, opened 2016-02-20
Fix for #290

kirill-konshin:
I also added a better handling of `eventHandlers.succeeded`.

### DIFF
--- a/src/Session.js
+++ b/src/Session.js
@@ -471,7 +471,7 @@ Session.prototype = {
   /**
    * Hold
    */
-  hold: function() {
+  hold: function(options) {
 
     if (this.status !== C.STATUS_WAITING_FOR_ACK && this.status !== C.STATUS_CONFIRMED) {
       throw new SIP.Exceptions.InvalidStateError(this.status);
@@ -497,27 +497,28 @@ Session.prototype = {
 
     this.onhold('local');
 
-    this.sendReinvite({
-      mangle: function(body){
+    options = options || {};
+    options.mangle = function(body){
 
-        // Don't receive media
-        // TODO - This will break for media streams with different directions.
-        if (!(/a=(sendrecv|sendonly|recvonly|inactive)/).test(body)) {
-          body = body.replace(/(m=[^\r]*\r\n)/g, '$1a=sendonly\r\n');
-        } else {
-          body = body.replace(/a=sendrecv\r\n/g, 'a=sendonly\r\n');
-          body = body.replace(/a=recvonly\r\n/g, 'a=inactive\r\n');
-        }
-
-        return body;
+      // Don't receive media
+      // TODO - This will break for media streams with different directions.
+      if (!(/a=(sendrecv|sendonly|recvonly|inactive)/).test(body)) {
+        body = body.replace(/(m=[^\r]*\r\n)/g, '$1a=sendonly\r\n');
+      } else {
+        body = body.replace(/a=sendrecv\r\n/g, 'a=sendonly\r\n');
+        body = body.replace(/a=recvonly\r\n/g, 'a=inactive\r\n');
       }
-    });
+
+      return body;
+    };
+
+    this.sendReinvite(options);
   },
 
   /**
    * Unhold
    */
-  unhold: function() {
+  unhold: function(options) {
 
     if (this.status !== C.STATUS_WAITING_FOR_ACK && this.status !== C.STATUS_CONFIRMED) {
       throw new SIP.Exceptions.InvalidStateError(this.status);
@@ -542,7 +543,7 @@ Session.prototype = {
 
     this.onunhold('local');
 
-    this.sendReinvite();
+    this.sendReinvite(options);
   },
 
   /**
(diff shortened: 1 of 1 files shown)


james-criscuolo:
Hi @kirill-konshin,
  I think this commit makes sense. I was a little thrown off because you mentioned in #290 that ideally promises would be returned, and that is not what this commit changes, but the changes here are certainly for the better and do fix your issue.

I'm open to accepting it, I'll just leave this open a few more days in case the other authors or anyone in the community has anything they'd like to see changed before accepting.

Thanks,
James

src/Session.js
```diff
@@ -612,15 +613,12 @@ Session.prototype = {
...
-        SIP.Timers.clearTimeout(self.timers.ackTimer);
-        SIP.Timers.clearTimeout(self.timers.invite2xxTimer);
-        self.status = C.STATUS_CONFIRMED;
-      };
-    }
+    this.reinviteSucceeded = function(){
+      SIP.Timers.clearTimeout(self.timers.ackTimer);
+      SIP.Timers.clearTimeout(self.timers.invite2xxTimer);
+      self.status = C.STATUS_CONFIRMED;
+      eventHandlers.succeeded && eventHandlers.succeeded.apply(this, arguments);
```
wakamoleguy: `eventHandlers.succeeded` should be pull off of the eventHandlers object immediately when `sendReinvite()` is called.  Otherwise, we run into issues when people reuse `options` objects across multiple calls.

wakamoleguy:
Do we use that `eventHandlers` option pattern anywhere else?  Seems a little odd.  I'm not terribly excited about furthering it.  `hold`/`unhold` are mostly just friendly, simple wrappers on top of `sendReinvite`, which could be used directly to handle a server response.  Since they are undocumented, it would *okay* to use this for them, but we should look to refactor that all away in the future.

One argument for adding this to `hold` and `unhold` instead of using a more complicated `sendReinvite` directly is that `hold` and `unhold` provide a pending action queue.  This commit, however, does not do anything to preserve options for actions in the queue.  If it did, would canceling of actions with different options be expected?

src/Session.js
kirill-konshin: Done

kirill-konshin:
I would say, that if `sendReinvite` have returned a promise -- this whole thing with callbacks will be unnecessary. Moreover, it's a very unsafe way:

```js
this.reinviteFailed = ...
```

Imagine user calls `sendReinvite` multiple times one by one with different handlers, what will happen?

I've tried to do as less changes as possible in order to solve a problem with uncontrolled `hold` and `unhold`.

Refactoring that involves Promises should be done everywhere, it's a big change, so I have planned to do it later on (with your permission).

kirill-konshin:
Any updates?

wakamoleguy:
We can probably merge it, but I'm still not crazy about it not working with the pending_actions queue.

kirill-konshin:
What do you mean "not working"? It works exactly the same way as before, the only difference -- options object is passed through hold/unhold, that's it.

wakamoleguy:
The options argument is only passed through if the hold happens immediately.  If the hold is queued due to pending actions, when the hold eventually takes place, no options argument is used.

kirill-konshin:
Wait, but success callback is assigned to `this.reinviteSucceeded` which will be executed later. By the way, there is no mechanism that handles multiple subsequent calls without overwriting `this.reinviteSucceeded` property.

wakamoleguy:
Sure there is.  If you call `hold` or `unhold` while `!this.isReadyToReinvite()`, the action will be placed in the pending_action queue without ever calling `sendReinvite()` at all.

kirill-konshin:
OK, I see.

Then my question would be -- what's the recommended way to determine whether hold/unhold went OK or failed? There are cases when nothing will be returned.

wakamoleguy:
We could add a new event, or set of events.  We have precedent for that type of thing with userMedia events.  A set of events could look like:

* `holdRequest` - When hold is called.
* `hold` - When the reinvite succeeds.
* `holdFailed` - When the reinvite fails.

The problem with this, of course, is that we already have a `hold` event.  We could change how that behaves, or instead make two new events `holdInviteSucceeded` and `holdInviteFailed`.  hold/unhold are not yet documented, so we do have some flexibility to repurpose the existing event.

Matching these up to individual hold calls would be left up the application, just like getUserMedia.  That is, if you call them multiple times, there is not necessarily information in the event to let you know which hold action it applies to.

It's certainly not perfect, but that may be a way to go about that is consistent with other pieces of the API.

kirill-konshin:
Events are a way to go, of course, but events don't give a clear understanding which particular hold/unhold was successful or failed.

I would say, that Promises + events is a cleaner way:

```
session.hold().then(function(){ ... }); // event holdInviteSucceeded is emitted before promise is resolved
```

If promises will be used consistently everywhere, then it will give a lot of benefits. Basically, any async method must return a Promise. But that's a big refactoring...

kirill-konshin:
When to expect a new version?

kirill-konshin:
Bump! Guys, please release a new version -- we get SIP.js from Bower and so far it's impossible to get anything from just master branch...

james-criscuolo:
Unfortunately, releasing a new version is not the simplest task for this project. We will look to do one soon.

james-criscuolo:
Version 0.7.4 is now available via bower. It will be available via npm later today, and on our website shortly.

PR #294, opened 2016-03-07
Allow overriding mediaHandlerFactory in UA#invite

markandrus:
Hi SIP.js team,

Would you please consider allowing `mediaHandlerFactory` to be overridden in `UA#invite`?

Thanks,
Mark


wakamoleguy:
I think this is likely harmless.  Some part of me thinks that the media handler factory should really be tied to the UA, and that if different media wants to be handled, a different agent should handle it.  But... I don't feel strongly about it, and the UA doesn't currently depend on it at all.

Concept ACK.

PR #296, opened 2016-03-11
Session emits "statusChanged" events

markandrus:
Hi SIP.js Team,

Currently, Transaction updates its `state` property via a `stateChanged` method which also emits a
"stateChanged" event. Would you consider a similar pattern for Session's `status` property?

This PR would add a `statusChanged` method to Session: all updates to `status` would be made through this method, and it would also emit a "statusChanged" event.

Thanks,
Mark

### DIFF
--- a/src/Session.js
+++ b/src/Session.js
@@ -26,7 +26,7 @@ var Session, InviteServerContext, InviteClientContext,
  *        (See the documentation for the mediaHandlerFactory argument of the UA constructor.)
  */
 Session = function (mediaHandlerFactory) {
-  this.status = C.STATUS_NULL;
+  this.statusChanged(C.STATUS_NULL);
   this.dialog = null;
   this.earlyDialogs = {};
   this.mediaHandlerFactory = mediaHandlerFactory || SIP.WebRTC.MediaHandler.defaultFactory;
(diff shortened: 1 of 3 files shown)


james-criscuolo:
Hi @markandrus,
  This and #294 look fine to be merged. I'm going to ping the other authors to give a look if they are interested, but otherwise I hope to merge both later this week.

Thanks,
James

wakamoleguy:
Are we okay giving the application control of execution in the middle of previously-atomic state changes?  Events are handled synchronously, and states are internal data, not public properties.  What's the desire to having these events?

markandrus:
@wakamoleguy my application has some scenarios (mostly due to forking) where it would be useful to know when/if we received the ACK to the INVITE/200 OK. For example, with this PR I could write

```js
function statusChanged() {
  if (session.status === C.STATUS_CONFIRMED) {
    // Do something now that we have received the ACK
  }
}

session.on('statusChanged', statusChanged);
```

I considered raising an event from `confirmSession` as an alternative approach, e.g. "confirmed", but I was not sure that I should invent something new.

james-criscuolo:
I think @wakamoleguy is right here, these states that we use are internal, so allowing people to make calls into the session mid status change could be potentially harmful. Your alternative seems preferable, creating a new event to be emitted when the ACK is received (and we are done doing any necessary processing).

Due to some other merges there are also conflicts with this PR. The first commit is still worth taking, but I don't think we would accept the 2nd, in favor of your alternate approach.

markandrus:
Thanks for the feeback @wakamoleguy and @james-criscuolo. I will prepare a new PR that adds the "confirmed" event instead.

josephfrazier:
I'm going to go ahead and close this, since there's no longer any action required. Looking forward to the new PR!

Issue #240, opened 2015-10-19
How to Override Websocket Implementation in 0.7.0+


lylepratt:
Hi Guys,

How do we override the WebSocket after SIP.js has already been instantiated? Overriding `window.WebSocket` in our dialer was working in SIP.js 0.6.4, however on 0.7.1, that no longer works.

Currently I'm instantiating SIP.js in a 3 second setTimeout with the `module.exports.WebSocket` hard coded to `cordova.plugins.phonertc.WebSocket`. However, obviously this is not a good solution. What is the correct way to do this?

We override the WebSocket implementation on iOS so we can support VoIP in the background. I was previously asking this question in a previous closed thread. Since this is applicable to 0.7.1 (and no 0.6.4) I decided to open a new one.

joseph-onsip:
Have you tried providing a custom `environment` as outlined in https://github.com/onsip/SIP.js/pull/108?

lylepratt:
Thanks for the tip. I'm might be wrong, but It appears that that would require me to compiling/configuring a custom version of SIP.js for each platform. I was trying avoid doing that because 1) We use Bower, (2) Its a pain. It's especially problematic when your WebSocket is a Cordova plugin (which is only instantiated AFTER the `deviceready` event is fired).

Is there no way to override WebSocket at runtime anymore...after SIP.js has been already instantiated?

Thanks for all you guys do!

joseph-onsip:
Oh, right... I had forgotten about the `deviceready` constraint. Unfortunately, if you load the Bower-provided version of SIP.js before that event is emitted, `SIP.Transport` already has a reference to the (nonexistent) `window.WebSocket`. Assuming you want to keep using the Bower version, perhaps you could only load SIP.js once `deviceready` has been emitted? I don't know if there's another good way around this at the moment.

JimGreenberg:
A brief update on this, the 0.11.0 release allows for the user to create and use a custom `Transport` similarly to how it is done with a custom sessionDescriptionHandler. There is no formal guide yet, but the [transport documentation](https://sipjs.com/api/0.11.0/transport) has more information. As always, our [mailing list](https://groups.google.com/forum/#!forum/sip_js) is there if you need help.

PR #299, opened 2016-03-15
Get global WebSocket on an on-demand basis

josephfrazier:
This makes it easier to use sip.js with a WebSocket implementation that
defines itself after sip.js has loaded (for example, one provided by a
Cordova plugin).

This should resolve https://github.com/onsip/SIP.js/issues/240. @lylepratt, could you verify that it works for you?


lylepratt:
@n9986 Can you try this in our new app? (would work in conjunction with https://github.com/BetterVoice/iOS-SIP.js

josephfrazier:
I'm closing this due to inactivity. Feel free to comment/reopen if there's any progress on the testing.

PR #305, opened 2016-03-29
Include REFER into allowed methods, fix #304

### LINKED ISSUE #304: INVITE request should include REFER as understood method in Allow header
Similarly to [the issue 270](https://github.com/onsip/SIP.js/issues/270), an INVITE request from unregistered UA is missing REFER in Allow header:

```
INVITE sip:username@example.com SIP/2.0
Via: SIP/2.0/WSS ws.example.com;branch=z9hG4bK3459694
Max-Forwards: 70
To: <sip:username@example.com>
From: "username" <sip:anonymous.9fipph@anonymous.invalid>;tag=jqcpja132o
Call-ID: 2gqlte3lmhrjb9t5lvbv
CSeq: 6621 INVITE
Contact: <sip:r794rcbd@ws.example.com;transport=ws;ob>
Allow: ACK,CANCEL,INVITE,MESSAGE,BYE,OPTIONS,INFO,NOTIFY
Content-Type: application/sdp
Supported: outbound
User-Agent: SIP.js/0.7.3
Content-Length: 5277
```

Again, [according to RFC](https://tools.ietf.org/html/rfc3261#section-20.5):

> All methods, including ACK and CANCEL, understood by the UA MUST be
   included in the list of methods in the Allow header field, when
   present.

growtofill:
Looks like a good first bug for me.

### DIFF
--- a/src/UA.js
+++ b/src/UA.js
@@ -30,7 +30,8 @@ var UA,
       'BYE',
       'OPTIONS',
       'INFO',
-      'NOTIFY'
+      'NOTIFY',
+      'REFER'
     ],
 
     ACCEPTED_BODY_TYPES: [


etamme:
As I commented on #304 I think that this should be dynamically added to the Allow header when a .on('refer') handler is created.  If there is no handler for the REFER method and we advertise support, what will the UA do when it receives a REFER?

growtofill:
> If there is no handler for the REFER method and we advertise support, what will the UA do when it receives a REFER?

[Send NOTIFY](https://github.com/onsip/SIP.js/blob/master/src/Session.js#L708) with the body of `SIP/2.0 603 Declined`.

> this should be dynamically added to the Allow header when a .on('refer') handler is created

I believe this approach is against the robustness principle. Additionally, it will require re-INVITE.

etamme:
It will **not** send a NOTIFY with a body of 603 decline.  It will send a SIP response with 603 Decline - a negative final response which would terminate the transaction for the REFER.  

I was unaware that the REFER handler is on the session rather than the UA, this obviously makes it more difficult to create a handler and dynamically build the Allow: header methods that are included in the INVITE request since the session does not exist.  

We are discussing the PR further internally and will follow up shortly.

growtofill:
> It will not send a NOTIFY with a body of 603 decline.

Why do you think so? It should pretty obvious from the code I've linked, here is the appropriate log:

```
Tue Mar 29 2016 17:45:34 GMT+0300 (EEST) | sip.transport | received WebSocket text message:

REFER sip:oodsbj2d@0.0.0.71:37637;transport=ws;ob SIP/2.0
Record-Route: <sip:0.0.0.79:5061;transport=tls;r2=on;lr;ftag=96946d9d6aee226a>
Record-Route: <sip:0.0.0.79;r2=on;lr;ftag=96946d9d6aee226a>
To: "username" <sip:anonymous.62dt19@anonymous.invalid>;tag=t0m22dobg2
From: <sip:1@example.com>;tag=96946d9d6aee226a
Contact: <sip:1@0.0.0.79:5061;transport=tls;did=b88.6d969753>
Call-ID: tj71pmq3qu51e604bhcu
CSeq: 2 REFER
Accept: application/sdp
Accept: application/media_control+xml
Accept: application/csta+xml
Accept: application/presentation_token_control+xml;q=1.0;m=video
Accept: message/sipfrag
Supported: norefersub
User-Agent: *
Via: SIP/2.0/WSS 0.0.0.71:10443;branch=z9hG4bK657a3ab23c8102181337d7748df9c630090414f2;rport
Via: SIP/2.0/TLS 0.0.0.79:5061;branch=z9hG4bKfb8d.021d0d7.0
Authorization: Digest username="1", realm="example.com", nonce="56fa960e000161b429f69386d77bd17a8e4259e8478bcdde", uri="sip:oodsbj2d@0.0.0.6;transport=tcp;did=b88.5fca2501", response="f5424ddfb5449d07f5515607059ea733", algorithm=MD5
Max-Forwards: 10
Refer-To: <sip:2@example.com>
Content-Length: 0



LoggerFactory.js:72 Tue Mar 29 2016 17:45:43 GMT+0300 (EEST) | sip.inviteclientcontext | REFER received
LoggerFactory.js:72 Tue Mar 29 2016 17:45:43 GMT+0300 (EEST) | sip.transport | sending WebSocket message:

SIP/2.0 202 Accepted
Via: SIP/2.0/WSS 0.0.0.71:10443;branch=z9hG4bK657a3ab23c8102181337d7748df9c630090414f2;rport
Via: SIP/2.0/TLS 0.0.0.79:5061;branch=z9hG4bKfb8d.021d0d7.0
To: "username" <sip:anonymous.62dt19@anonymous.invalid>;tag=t0m22dobg2
From: <sip:1@example.com>;tag=96946d9d6aee226a
Call-ID: tj71pmq3qu51e604bhcu
CSeq: 2 REFER
Supported: outbound
User-Agent: SIP.js/0.7.3
Content-Length: 0



LoggerFactory.js:72 Tue Mar 29 2016 17:45:43 GMT+0300 (EEST) | sip.transport | sending WebSocket message:

NOTIFY sip:1@0.0.0.79:5061;transport=tls;did=b88.6d969753 SIP/2.0
Route: <sip:333c9357db@0.0.0.71:10443;transport=wss;lr;ovid=1a886ecf>
Route: <sip:0.0.0.71:5061;transport=tls;lr;ovid=1a886ecf>
Via: SIP/2.0/WSS ws.example.com;branch=z9hG4bK3287965
Max-Forwards: 70
To: <sip:1@example.com>;tag=96946d9d6aee226a
From: "username" <sip:anonymous.62dt19@anonymous.invalid>;tag=t0m22dobg2
Call-ID: tj71pmq3qu51e604bhcu
CSeq: 2272 NOTIFY
Event: refer
Subscription-State: terminated
Content-Type: message/sipfrag
Supported: outbound
User-Agent: SIP.js/0.7.3
Content-Length: 20

SIP/2.0 603 Declined
```

etamme:
I believe this is a bug - if there is no handler setup, the UA should respond with a 4XX-6XX response as indicated by https://tools.ietf.org/html/rfc3515#section-2.4.2.  

A NOTIFY with a reason code in the body should only be sent if the UA has accepted the REFER and is in a pending state of processing it.

james-criscuolo:
There are 2 other changes I'm about to make as a result of this:

- the 603 Decline in the notify should be a 503 Service Unavailable

- we should actually be 603 Declining

Thanks for the PR!

fredguth:
Couldn't find `followRefer (callback)` in the docs.

PR #310, opened 2016-04-12
Move releases to master branch, use `npm version`

josephfrazier:
[Issue 291] reminded me how much of a pain it is to prepare and publish
a new release, so I thought I'd try out something different, inspired by
[webtorrent]. Instead of having a separate release branch, let's just
keep releases on master. This avoids all the extra git stuff required to
do it the old way, while still supporting bower users by checking in
`dist/` files upon release.

With these changes, preparing a release is as simple as

    npm version patch # patch/minor/major/etc

Using the [scripts] in package.json, this will:

* Remove old `dist/` files
* Reinstall `node_modules` from scratch
* Do a clean build and test
* Add new `dist/` files
* Bump version numbers in `package.json` and `bower.json`
* Commit and tag the release

This result will be a single tagged commit with all the changes in
`dist/`, `package.json`, and `bower.json`.

After that, all that's needed to publish the release is:

    npm publish # for npm
    git push --follow-tags # for bower

`npm publish` creates a new commit that removes the `dist/` directory,
so that it can remain ignored until the next release.

[Issue 291]: https://github.com/onsip/SIP.js/pull/291#issuecomment-208897713
[webtorrent]: https://github.com/feross/webtorrent
[scripts]: https://docs.npmjs.com/cli/version

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,25 +1,7 @@
 # How To version release SIP.js
 
-(These are developer notes.)
-
-(These may not be entirely accurate.)
-
-(Eric Green demands credit for these.)
-
-(So remember that if it's broken, you know who to contact.)
-
-(If you don't know, it's Eric Green.)
-
-* On your own github, checkout last tagged release on a new branch (note: this can be done on the repo's release branch, instead of making your own)
-* remove all dist files
-* cherry pick commits you want using -x flag (for "hot patch" releases)
-* once ready, test.
-* update version number on master
-* cherry pick version number commit to new branch (or just merge master, if you want everything)
-* build and test.
-* test again
-* add new dist files (git add -f if it complains)
-* commit
+* `git checkout master && git pull`
+* `npm version patch # patch/minor/major`
 * test npm:
 
     ```shell
@@ -68,11 +50,7 @@
     set +e
     ```
 
-* push to local github
-* merge (this step and the above one can be skipped if you just do it on the the repo's release branch itself)
-* git tag (your version number)
-* git push --tags
-* get a clean release (as in, fresh clone)
-* npm publish
+* `npm publish`
+* `git push --follow-tags`
 * do release notes on github and release!
 * update website
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-*.min.js binary
+dist/* binary
--- a/package.json
+++ b/package.json
@@ -46,7 +46,11 @@
   "scripts": {
     "repl": "beefy test/repl.js --open",
     "build": "grunt build",
+    "version:build": "git clean -fxd && npm install && grunt && git add -f dist",
+    "version:bower": "node ./scripts/version_bower.js && git add bower.json",
+    "version": "npm run version:build && npm run version:bower",
     "prepublish": "cd src/Grammar && mkdir -p dist && pegjs --extra-options-file peg.json src/Grammar.pegjs dist/Grammar.js",
+    "postpublish": "git rm -rf dist && git commit -am 'Remove dist/ files'",
     "postinstall": "[ -f src/Grammar/dist/Grammar.js ] || npm run-script prepublish",
     "test": "grunt travis --verbose"
   },
--- a/scripts/version_bower.js
+++ b/scripts/version_bower.js
@@ -1,0 +1,10 @@
+#!/usr/bin/env node
+"use strict";
+
+// This is run during `npm version`, to update the version in bower.json
+// See https://github.com/onsip/SIP.js/pull/310#issuecomment-209517244
+var fs = require("fs");
+var npm_version = JSON.parse(fs.readFileSync("package.json")).version;
+var bower_json = JSON.parse(fs.readFileSync("bower.json"));
+bower_json.version = npm_version;
+fs.writeFileSync("bower.json", JSON.stringify(bower_json, null, 2));


josephfrazier:
https://github.com/josephfrazier/SIP.js/commits/no-release-branch-test shows an example of what the commits look like.

RELEASE.md
```diff
@@ -68,11 +50,7 @@
...
     ```
 
-* push to local github
-* merge (this step and the above one can be skipped if you just do it on the the repo's release branch itself)
-* git tag (your version number)
-* git push --tags
-* get a clean release (as in, fresh clone)
-* npm publish
+* `npm publish`
+* `git push --follow-tags`
```
dbmikus: Being pedantic, but I would imagine we want this line and the line above to be swapped. So, instead

```
git push --follow-tags
npm publish
```

james-criscuolo:
I think this PR covers several different issues, and I at least need some clarification before I'd be ok merging it.

- The release branch, while I don't particularly care for it, serves a good purpose. Remember during the `0.6.x` days when promises got onto `master` and we needed more releases, but couldn't include the promise commits? For a lot of releases, we just cherry-pick the entire `master` branch over, but there are valid cases where that isn't necessarily the case. How does the new plan look to tackle that?

- This is more philosophical I guess, but why were the `test`, `build and test`, and `test again.` lines removed from `release.md`? Part of the release process is extensive testing, which was the real time-consuming portion, not the git movement that the file used to lay out for us. The new system seems to promote more common releases, which we don't necessarily want.

- Why `versiony`? I don't believe a dependency should ever be added without being defended properly. In this case, combined with the scripts, it seems to save us changing two characters for some, but not all, releases. This also moves more towards the idea of more common releases.

dbmikus:
`versiony` seems really lightweight (no other dependencies), but for our purpose we could have just used:

```
var fs = require("fs");
var npm_version = JSON.parse(fs.readFileSync("package.json")).version;
var bower_json = JSON.parse(fs.readFileSync("bower.json"));
bower_json.version = npm_version;
fs.writeFileSync("bower.json", JSON.stringify(bower_json, null, 2));
```

Whether it is justified or not, my paranoia of NPM has grown since the leftpad debacle. For very simple code, I prefer to just do it myself.

RELEASE.md
josephfrazier: I originally had it that way, but then I added in the `postpublish` hook to clean up `dist/`, so I swapped the command order so the cleanup commit would be pushed as well. I could change it back, then add another push after publish, though:

```bash
git push --follow-tags
npm publish
git push
```

josephfrazier:
Thanks for the feedback, I'll see if I can address your concerns:

> The release branch, while I don't particularly care for it, serves a good purpose. ...

That's a good point, @james-criscuolo. These changes don't necessarily require that we stop using the `release` branch as a place to stage releases outside master. We could still cherry-pick commits from `master` to `release`, then do the release process itself with `npm version patch`, etc.

> why were the `test`, `build and test`, and `test again`. lines removed from release.md?

The "version:bower" script that is run by `npm version patch` includes the automated tests via `grunt`, so those aren't getting omitted. I left the "test npm" and "test bower" items in `RELEASE.md`, since they provide actual smoke tests in real browsers (rather than mocked-up tests that run in PhantomJS), and those items still include `# more tests` comments, for what that's worth. I think there's a lot of room for improvement here in general (say, by having Travis run integration tests in browsers), but in the mean time, I'd be willing to add back in another list item with a bit more emphasis on manual testing, if you have preferred wording for it, @james-criscuolo.

> `versiony` seems really lightweight (no other dependencies), but for our purpose we could have just used: ...

Good call, @dbmikus. Originally, I was concerned that `versiony` might do some extra processing to avoid reordering JSON keys when parsed/stringified, but that appears not to be the case, so I've updated this PR to include a `scripts/version_bower.js` file that is used instead of `versiony`.

> Why `versiony`? I don't believe a dependency should ever be added without being defended properly. In this case, combined with the scripts, it seems to save us changing two characters for some, but not all, releases. This also moves more towards the idea of more common releases.

As noted above, I've now removed `versiony` from the PR. However, doing this in an automated way makes it easier and more fool-proof to do releases. Having manual steps like "update bower.json" only makes it harder to carry out a release, and uses energy/mindshare that would be better spent on, for example, manual testing of the code to be released. In my opinion, the mechanistic parts of the release process shouldn't be the ones we spend our time on. Let's leave those to the machines.

Along a similar line of thinking, having more frequent releases makes it easier/faster for users to take advantage of recent changes, without having to do things like maintain a fork or [`npm install` straight from Github](https://github.com/onsip/SIP.js/pull/312#issue-148626663).

james-criscuolo:
I haven't looked into too many of the details, but if the `release` branch is sticking around, at least a few of the lines removed from `release.md` need to be reintroduced. Version still needs to be updated on `master`, then cherry-picked/rebased with the rest of the commits. I also don't really know what the local branch release stuff was about, but it doesn't seem to necessarily be obsoleted by this, but someone who has done a release in that fashion can probably explain if it's still necessary.

As for `test` wording, the lines as they were seem fine.

wakamoleguy:
Late to the party, but I'd definitely agree on keeping the release branch.  SIP.js has followed [this branching model](http://nvie.com/posts/a-successful-git-branching-model/) (where our `master` is their `develop` and our `release` is their `master`). I think deviations from that should be *very* heavily scrutinized.  Just my two cents.

My only other fear would be automating to the point where mistaken releases are easy.  It looks like there are still separate prep and publish steps, so that's probably fine.  If it goes too far in that direction, maybe additional 'Press Enter to Continue' prompts would go a long way.

Issue #276, opened 2016-01-22
mediaHandlingFactory for testing calls with node


andrewvmail:
Hi Guys, just wanna say wonderful library you guys have here. 

Now that said, I know that nodejs support is pretty experimental. But could someone give me guidance on how to get mediaHandling working something in the line of accepting the calls and playing an audio of some sort and hang up. 

I was able to get everything going, overriding the mediaHandlerFactory with an empty get and set descriptions functions. But not sure where to start to get media flowing from both UA running in node.

Thanks
Andrew

josephfrazier:
It used to be the case that you could use [`wrtc`](https://github.com/js-platform/node-webrtc) to receive audio, after providing it as the WebRTC implementation as described in https://github.com/onsip/SIP.js/pull/108. However, `wrtc` dropped MediaStream support somewhere around https://github.com/js-platform/node-webrtc/commit/992543f249dca33060eef76090c4aaebd9f24985, so you'd have to use an old version to make that work.

I'm not aware of any other MediaStream implementations for Node, but you might be able to use something like [NW.js](http://nwjs.io/) or [Electron](http://electron.atom.io/) to create a standalone webapp that supports media streaming, if you're interested in that.

josephfrazier:
I'm closing this for now. https://github.com/feross/webtorrent-hybrid/issues/5#issuecomment-178995442 lists some libraries that may be useful for server-side MediaStream support. Specifically, if [electron-webrtc](https://github.com/mappum/electron-webrtc) adds MediaStream support, it may be possible to use it to receive (and potentially send?) audio.

PR #313, opened 2016-04-15
add supports for react-native

zxcpoiu:
add new environment to support react native.

the webrtc library use https://github.com/oney/react-native-webrtc
and websocket uses react-native's polyfill.

### DIFF
--- a/src/environment_react_native.js
+++ b/src/environment_react_native.js
@@ -1,0 +1,38 @@
+"use strict";
+var DeviceEventEmitter = require('react-native').DeviceEventEmitter;
+var WebRTC = require('react-native-webrtc');
+
+var toplevel = global.window || global;
+
+module.exports = {
+  WebSocket: toplevel.WebSocket || require('ws'),
+  Transport: require('./Transport'),
+  open: toplevel.open, // null: only use in followRefer() in Session.js
+  Promise: toplevel.Promise || require('promiscuous'),
+  timers: {
+    'setTimeout': toplevel.setTimeout,
+    'clearTimeout': toplevel.clearTimeout,
+    'setInterval': toplevel.setInterval,
+    'clearInterval': toplevel.clearInterval
+  },
+
+  // Console is not defined in ECMAScript, so just in case...
+  console: toplevel.console || {
+    debug: function () {},
+    log: function () {},
+    warn: function () {},
+    error: function () {}
+  },
+
+  MediaStream: WebRTC.MediaStream,
+  getUserMedia: WebRTC.getUserMedia,
+  RTCPeerConnection: WebRTC.RTCPeerConnection,
+  RTCSessionDescription: WebRTC.RTCSessionDescription,
+
+  // we don't need html audio/video tag to render media.
+  addEventListener: DeviceEventEmitter.addEventListener, // only used in UA.js that try to listen 'unload' browser event
+  HTMLMediaElement: toplevel.HTMLMediaElement, // null: only used in desugar() in Session.js.
+  attachMediaStream: toplevel.attachMediaStream, // null: only used in WebRTC/MediaStreamManager, and has a fall back function.
+  createObjectURL: toplevel.URL && toplevel.URL.createObjectURL,
+  revokeObjectURL: toplevel.URL && toplevel.URL.revokeObjectURL
+};
--- a/package.json
+++ b/package.json
@@ -7,6 +7,9 @@
   "browser": {
     "./src/environment.js": "./src/environment_browser.js"
   },
+  "react-native": {
+    "./src/environment.js": "./src/environment_react_native.js"
+  },
   "homepage": "http://sipjs.com",
   "author": "OnSIP <developer@onsip.com> (http://sipjs.com/authors/)",
   "contributors": [


zxcpoiu:
Hmm...
I don't know why Travis build failed.

The `react-native` column in package.json is used to tell react native packager.

james-criscuolo:
Hi @zxcpoiu,
  Once the Travis check is passing for this PR and the later one (#314) we will consider them.

Thanks,
James

zxcpoiu:
yeah, I will do that, thanks!

src/environment_react_native.js
```diff
@@ -0,0 +1,38 @@
+"use strict";
+var DeviceEventEmitter = require('react-native').DeviceEventEmitter;
+var WebRTC = require('react-native-webrtc');
```
josephfrazier: I'm not familiar with the react-native workflow, but I'm assuming that it automatically provides the `react-native` dependency. Is that the case?

What about the `react-native-webrtc` dependency? I'm guessing it doesn't get automatically provided by react-native, so is the assumption that the developer will have to manually install it alongside `sip.js`?

josephfrazier:
React Native support is an exciting prospect, but I'm curious whether this is the best way to go about it, as you can see in my code comment above.

For what it's worth, you should be able to add React Native support at runtime as outlined in https://github.com/onsip/SIP.js/pull/108, and you could even create a separate package (say, `react-native-sip.js`) that wraps everything up cleanly into a single module.

src/environment_react_native.js
zxcpoiu: yes, if you are running on react native, it will provide:

* react-native module
* global var
* global WebSocket and Promise
* the rest except I commented as `null`

`react-native-webrtc` is a third party module needs to install manually.

the field `react-native` in package.json is specifically for react native, it's packager will check that field and replace files like `browser` field does.

zxcpoiu:
a different module is a way to go, but it looks a bit odd which just have a index.js wrapper and a package.json to specify `sip. js` and `react-native-webrtc` deps and also harder to maintain issues.

the think I'm not sure about is it lacks support of `global.open` function which used in `follow_refer()`

does node user have to install `wrtc` manually? if yes, I don't see any bad would happen. it won't touch neither `environment_browser.js` nor `environment. js` because the `react-native` field in `package. json`.

Sorry, I have limited knowledge about these.
How about we merged it as a `experiment` comment and I can provide a README section?

josephfrazier:
> the think I'm not sure about is it lacks support of global.open function which used in follow_refer()

That's not really a big deal, as [it's used only when the Refer-To URI is not sip: or sips:](https://github.com/onsip/SIP.js/blob/fa6e596ee9998780f01ee5eaf55e7b75a9b16860/src/Session.js#L257-L265), which is rather uncommon.

> does node user have to install wrtc manually?

One of the server-side sip.js use cases is: for signaling only, without any media handling. That, along with undeveloped media-handling support in Node.js, is partially why sip.js doesn't depend on server-side media module. In fact, [`wrtc` dropped MediaStream support](https://github.com/onsip/SIP.js/issues/276#issuecomment-173942462), so it doesn't even work for audio anymore (and I had only ever gotten it to work for receiving audio anyway). At any rate, I don't think it would be appropriate to list as a dependency and force users to download, especially since most of our users are browser-based. Of course, this is also the case with `react-native-webrtc` (especially given its size). I was really just more curious about how all the dependencies were handled. I think it's totally appropriate to ask users to install `react-native-webrtc` manually if these changes were merged in.

> a different module is a way to go, but it looks a bit odd which just have a index.js wrapper and a package.json to specify sip. js and react-native-webrtc deps and also harder to maintain issues.

I don't think wrapper modules like this are all that uncommon (AngularJS wrappers for various modules come to mind, as do React Native wrappers, for that matter). Whether they're a good idea is a separate question, of course. You have a good point about issue maintenance, for one. I can see how it would be a pain to have to "forward" sip.js issues to this repo, while keeping the react-native ones elsewhere.

On the other hand, supported another platform like this implies that other developers would need to do extra testing to avoid accidentally breaking compatibility with react-native. Given the small scope of this change, I imagine that's relatively unlikely, but I could definitely imagine someone changing `environment_browser.js` (from which `environment.js` extends), and forgetting about  `environment_react_native.js`. You could reduce this possibility by having `environment_react_native.js` also extend from one of the other two `environment` files.

> Sorry, I have limited knowledge about these.

No problem, I've been learning things from this thread too 😄 

> How about we merged it as a experiment comment and I can provide a README section?

I think that adding a section to the README demonstrating how to set up a "Hello World" project would definitely help here, as it would allow the repo collaborators to try it out and see how it works, as well as guiding future users, and providing a reminder for why the react-native code is here in the first place.

Personally, I'm curious to see how this would work out, and I'd be excited to give it a try if you provided instructions (even better, a premade "Hello World" project I could clone and run). I hope this helps!

zxcpoiu:
you are right, this may cause some confusions and add complexity for maintaining.

and yes, to have hardcoded dependencies of an specific implementation doesn't feel right.
and it seems there are no way for user to specify a dependence via tag when `npm install` currently.

`react-native-webrtc` is the only one implementation on `react-native` AFAIK, and there is another implementation for mobile webrtc called [openwebrtc](https://github.com/EricssonResearch/openwebrtc) by Ericsson. it can run on `Cordova` but has not porting to `react-native` yet. maybe I will do that when I have time to experiment.

>You could reduce this possibility by having environment_react_native.js also extend from one of the other two environment files.

I am not exactly understand what you mean about `extend` from one of the other two.

doesn't it `environment_react_native.js` replaced `environment.js` if we have field  `react-native` in `package.json`?

like `environment_browser.js`, how can I extend from it if I'm the replacer environment? 😕 

anyway, it is not an easy work for sure.
especially react-native is in it‘s youth and it is commonly to have to manually tweaking your setup (android/iOS library paths, permissions and so on) after `npm install`, not to mention you don't have `audio/video` tag like in browser, so you need to control media in `react-native` way. 😵 

FWIW, there is `rnpm`, it is a npm for `react-native `, maybe it helps.

alright, let's leave this PR for a while here for further discussion.

I will make an easy repo for you guys to play around and see what we will get.

in the mean time, I will eat more documens and try to level-up in order to discuss with you. 😱 

thanks for your time and patiences and of course all the decent works.

alioguzhan:
@zxcpoiu Hey, can i use sip.js with react native now? Or we should wait till your PR gets merged?

In the mean time, any suggestion d be great for implementing sip with react native.

Thanks

zxcpoiu:
@alioguzhan yes of course, `sip.js` is mature enough, but `react-native` still stands on the blooding-edge. My scenario is audio only, and reaching to semi-production stage, but still figuring out what is the best practice and some unclear issues to be fixed. I will create a demo once these things are done.

**how to use:**

once you have all `react-native` modules setup, just **manually patch this PR**, and you can use `sip.js`.
you may want to check/patch #314 as well.

**below is some point I can tell:**

* If you want to use video, you may have lot more pains than audio only, since you need to dealing with view rendering.
* the connection-handling / life-cycle in mobile world is much more painful than in PC's browser.
* you may need to adjust lots of logic flow **depends on your scenario**, this is [my dirty works](https://github.com/zxcpoiu/SIP.js/commits/react-native)
* you would need a stable sip/turn server compatible enough with `sip.js`

alioguzhan:
@zxcpoiu Thanks for detailed explanation! Looks like it is almost impossible using sip.js with proper and native ways. But what about webviews?

What if we put a webview includes only <video> and some <button> elements. I can write all logic in a html file with some javascript. Can we make it work? Sure, i will try this. Asking just for a brainstorm :)

zxcpoiu:
oh, sorry for the misleading.

`sip.js` is mature enough and handled sip signals well, the problems is in `native side` ( `react-native`/`react-native-webrtc`/native limitations ) no matter used in webview or native view.

The point is all based on what features you want to implement.

if you want just a "click to call" feature, then it is fairly easy to implement, in webview/chrome/react-native.

but if to refers to "receiving calls"/"support audio routing to vary devices"/"permissions and system call priority"/"push notification wake up", then it would be complex. ( that's my case )

because PC's browser doesn't have this conflict environments and mostly handles all for you well.

so, `sip.js` itself is absolutely can be used in mobile without problems, the blooding part is `react-native` and `react-native-webrtc` which acts differently than PC's browser. you don't have to worry about `sip.js` much.

I was successfully got `sip.js` running in android chrome with pure html and js, but the limitations of native accessibility forced me to go with native app.

I don't have experiences in `webview`, but I think it's worthy a try if you just have to implement something like `click to call`

my two cents.

alioguzhan:
@zxcpoiu got it. I need exactly same features as in desktop/browser environments. Like receiving and sending video&audio calls based on users' sip account credentials. Since my app will be a launch screen app, I don't interested about wake on push notifications part.  Because app will be always open. 

Looks like i need to implement one of those native sip libraries written in Java. And try to use them as native modules from react native side. 

Can you suggest anything else?

zxcpoiu:
> Because app will be always open.

I have doubt with it.

Both Google and Apple tends to put more and more restrictions on persistent connections and clearly recommends that voip apps should use push notification to wake app up.
The app live or not, the decision all owned by the system, eventually. But it is only my bet based on my observation.

If you have to use `react-native`, currently, your best bet is: `sip.js` + `react-native-webrtc` (of course, or I won't use them )

**you need to decide:**

* you want to handle sip signaling in `js` or `native`
* use persistent connection to register or wake up by push notification
* license
* do you really need react-native

**for another options:**

* ( signal functions on js, media functions on native ) porting [openwebrtc](https://github.com/EricssonResearch/openwebrtc) to a `react-native` module. but you still need `sip.js` to handle sip signals in js.

* ( both signal and media functions on native ) porting [linphone](https://www.linphone.org/) to react-native 

* ( both signal and media functions on native ) porting [pjsip](http://www.pjsip.org/) to react-native  there is a repo [react-native-pjsip](https://github.com/datso/react-native-pjsip) tring to porti from pjsip, but still in development.


that's what I can tell. and this topic is way beyond the scope of `sip.js`
it's never been easy to make a decent sip app, good luck

rcastro83:
zxcpoiu

I have a doubt on the options passed to the accept method of InviteClientContext:

with this object
{
                media: {
                    constraints: {
                        audio: true,
                        video: false,
                    },
                    render: {
                    	local: this.containerLocal,
                    	remote: this.containerRemote,
                    }
                },
            };

Where the local and remote are ref's to <RTCView> .

I have to do this ?

Because on this scenario i'm not getting audio and asterisk place the call on hold because of this "c=IN IP4 0.0.0.0"

Any idea ? 

Thanks.

zxcpoiu:
There is no `video` nor `audio` elements in `react-native` like in browser. ( maybe there are some modules   do the polyfill that I didn't notice.  )
This different behavior is one of the main reason that this PR can't be merged smoothly.

`react-native-webrtc` use different way to render media stream,

for audio, controlled via Stream.getTrack() directly, and you don't have to render on a `audio` element, it's done by webrtc itself automatically.

for video, `react-native-webrtc` use a `react native View` to render, you can check the source code for the implementation.

zxcpoiu:
about questions to `react-native-webrtc`, please go to https://github.com/oney/react-native-webrtc for helps.

james-criscuolo:
At this point, I feel this is more of a discussion topic than something we will accept as a PR.  `react-native-webrtc` is also a very active project, so it's likely what's necessary has changed since the time this PR was opened.

Please feel free to continue this as an issue or in the google group.

Thanks,
James

ostrichegret:
@zxcpoiu Hi, I am very interested on React Native integration and want to try React Native + React Native Webrtc + SIP.js

So i did this:
react-native init WebrtcDemo
cd WebrtcDemo
npm install
npm install --save react-native-webrtc
open package.json, add dependency "sip.js": "https://github.com/zxcpoiu/SIP.js.git" for react native mod
npm install
open App.js, add these lines:
import {
  RTCPeerConnection,
  RTCMediaStream,
  RTCIceCandidate,
  RTCSessionDescription,
  RTCView,
  MediaStreamTrack,
  getUserMedia,
} from 'react-native-webrtc';

import sip from 'sip.js';

and i got error: bundling failed: UnableToResolveError: Unable to resolve module `events` from `D:\WebrtcDemo\node_modules\sip.js\src\EventEmitter.js`: Module does not exist in the module map.

Do you have a demo to do sip call? 
Thank you very much

PR #314, opened 2016-04-15
add ice into candidate if not in sdp.

zxcpoiu:
some WebRTC implementation may not include ice candidate after createOffer/createAnswer.

Since sip.js doesn't support ice trickling and using defer to wait all candidates to be found, we need to make sure our sdp has ice candidates.

It would be better to make it optional and default enabled for future support of ice trickling.

### DIFF
--- a/src/WebRTC/MediaHandler.js
+++ b/src/WebRTC/MediaHandler.js
@@ -494,8 +494,19 @@ MediaHandler.prototype = Object.create(SIP.MediaHandler.prototype, {
         }
         return deferred.promise;
       })
-      .then(function readySuccess () {
-        var sdp = pc.localDescription.sdp;
+      .then(function shouldAppendIceToSdp() {
+        // TODO: should make this optional if we want to support ICE Trickling
+        if (pc.localDescription.sdp.indexOf('candidate') < 0) {
+          return SIP.Utils.promisify(pc, methodName, true)(constraints)
+            .then(function(sdp){
+              return sdp.sdp;
+            });
+        } else {
+          return '';
+        }
+      })
+      .then(function readySuccess (sdp) {
+        sdp = sdp || pc.localDescription.sdp;
 
         sdp = SIP.Hacks.Chrome.needsExplicitlyInactiveSDP(sdp);
         sdp = SIP.Hacks.AllBrowsers.unmaskDtls(sdp);


zxcpoiu:
updated! tell me if I did anything wrong. :)

james-criscuolo:
Hi @zxcpoiu,
  Does the creation of the offer/answer on line 486 not cause this inclusion of the ice candidates? I don't doubt that this fixed a problem for you, I just want to make sure we know exactly why this is happening. What implementation did this affect?

As the PR stands, it looks like you may create duplicate offers/answers, so I'm not sure why the second one makes a difference. Maybe the ice states have changed since we last looked at this code? Also, the code you added does not wait for any particular ice state to occur, so at minimum that would need to be added as well.

- James

zxcpoiu:
sorry for the late,

in [line L491](https://github.com/zxcpoiu/SIP.js/blob/option-get-ice/src/WebRTC/MediaHandler.js#L491), doesn't it defer the promise and wait until `iceGatheringState`  changed to `complete`? or did I missed something.

I'm using `react-native-webrtc` implementation, and the core webrtc library also using google one.

chrome and firefox seems includes all candidates into sdp by default. 
I'm not sure if `includes all candidates into sdp` is the default action of webrtc core library or it is a customization by chromium and firefox, my guess is later because google's webrtc support ice trickle by default.

**All my assumption needs to be verified carefully. I will keep experiment**

so that's why I need to issue `createOffer/createAnswer` again after the state become `complete` ( it will cache ice candidate, and bundle it into sdp if any. )

there is an article says: 
> Firefox includes the Candidate in the Offer SDP.

https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/WebRTC_basics

james-criscuolo:
Line L491 checks to make sure that the ice state has not already switched to complete before we arrived at that point. The else condition under it is the piece that waits until the `complete` state.

I think the correct way to handle this would be to call createOffer/Answer once, either where it currently is, or where you are suggesting, but not both. Your research seems to point to your createOffer/Answer call being in the correct position, but as you stated, we'd definitely need sufficient evidence (both through your testing and ours) that this will still work.

Intuitively, your placement seems more correct, given that the candidates do need to be in the sdp that's being generated, so hopefully some testing will back that intuition. Let me know if you need help factoring out the original createOffer/Answer call, and I can look into providing a patch.

Thanks,
James

zxcpoiu:
thank you James
will ping you if I got any result.
my direction is to read the webrtc source code, for the result be much more solid.

belows are some notes

**Some points noted here for who are interested in this**

* the `ice gathering process` triggered after [setLocalDescription()](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-12#section-4.1.5) is called.

* we must considering support of ice trickling as an option sooner or later.

**My observation and assumption so far: ( NEED VERIFIED )**

1. google's webrtc library enable ice trickling by default, so it does not bundle candidates into sdp.

2. Firefox does bundle ice candidates into sdp

3. Chrome does not bundle original, but it does now.


**Links**

[setLocalDescription starts candidate gathering?](https://groups.google.com/forum/#!topic/discuss-webrtc/W412UnSooCc) and [necessary to send ice candidates across, or do they come packaged in offer/answer data?](https://groups.google.com/forum/#!topic/discuss-webrtc/UOnopWJ1l44) describes the same issue which ice candidates are not included into sdp about chrome.  

[Disable ICE Trickling](http://muaz-khan.blogspot.tw/2015/01/disable-ice-trickling.html) this blog describes how to disable and workaround of trickle ice.  
[WebRTC ICE candidate in Firefox](http://stackoverflow.com/questions/25488318/webrtc-ice-candidate-in-firefox) is a related SO thread.

markandrus:
WebRTC implementations [should include any ICE candidates gathered in the `localDescription`](http://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-currentlocaldescription).

> The `currentLocalDescription` attribute MUST return the last value that algorithms in this specification set it to, completed with any local candidates that have been generated by the ICE Agent since the offer or answer was created. Prior to being set, it returns null.

This is also specified in [JSEP](https://tools.ietf.org/html/draft-ietf-rtcweb-jsep-14#section-4.1.7):

> The currentLocalDescription method returns a copy of the current negotiated local description - i.e., the local description from the last successful offer/answer exchange - in addition to any local candidates that have been generated by the ICE Agent since the local description was set.

Granted these are drafts, but Chrome and Firefox do this. Perhaps `react-native-webrtc` is doing something differently?

zxcpoiu:
Thanks @markandrus for your tips.

`react-native-webrtc` is simply a wrapper of native webrtc library ( same library woth chrome, but in different branch version ) as I see now.

if I call `createOffer` again after ice completed, the returned sdp contains all candidates, it means even in `react-native-webrtc` 's native webrtc lib would bundle candidates into sdp if any.

so the question would became:

* when should update pc.localDescription
  1. update pc.localDescription when got a new ice candidates?
  2. stores all candidates in ice agent, bundle into sdp only when user access pc.localDescription

* how to insert different media type candidates into pc.localDescription at a proper position respect to video/audio parts
  
Hmm... I guess this discussion is a little bit off topic and out of scope of this repo.

This really should dig into source code, to see how native webrtc acts, and how chrome's abstraction api does. I'm still too young to comment on this. :(

james-criscuolo:
Hi,
  As mentioned above, this PR got a little out of scope. As with #313, please continue this discussion as an issue or in our Google Group.

Thanks,
James

PR #321, opened 2016-04-27
Session emits "confirmed" with the ACK

markandrus:
Hi SIP.js Team,

Here is the followup to #296. I included the first commit, "checkTransaction should use stateChanged", because @james-criscuolo indicated it would be worth taking; however, the real change though is in the second commit, which updates Session to emit a "confirmed" event with the ACK it receives.

Thanks,
Mark


james-criscuolo:
Hi @markandrus,
  This looks good, I'll let the other authors know to give a look before merging sometime next week.

Thanks,
James

Issue #323, opened 2016-04-29
Notification about INFO message was recently removed


alexandrkozlov:
Hi,

You recently removed notification about INFO message called from _Session.js_:
>      case SIP.C.INFO:
>        if(this.status === C.STATUS_CONFIRMED || this.status === C.STATUS_WAITING_FOR_ACK) {
>               ..........
>                this.emit('info', request)
>               ..........
>                request.reply(200);

We really need it in our application, please replace:
>      case SIP.C.INFO:
>               ..........
>                request.reply(415, null, ["Accept: application/dtmf-relay"]);

with the following 
>      case SIP.C.INFO:
>               ..........
>              if (!this.emit('info', request))
>                request.reply(415, null, ["Accept: application/dtmf-relay"]);
>              else
>                request.reply(200);

josephfrazier:
Hi @alexandrkozlov, can you link to the commit in question? I'm having trouble finding where `emit('info', request)` was _ever_ called anywhere.

alexandrkozlov:
Sorry, right I'm also not sure that it was really called before (maybe it was our temporary local changes)

So, please replace line 694 of https://github.com/onsip/SIP.js/blob/master/src/Session.js with 
>if (!this.emit('info', request))
>           request.reply(415, null, ["Accept: application/dtmf-relay"]);
>         else
>           request.reply(200);

Probably the best way is to send 200 OK only if listener really process INFO, but anyway it's better than right now

alexandrkozlov:
Let's continue in #324

PR #324, opened 2016-04-29
Allow overriding INFO handler

markandrus:
This addresses #323 by allowing you to specify an `infoHandler' when calling `UA#invite` or `Session#accept`.


markandrus:
A couple questions:

* I am not sure if this parameter should be called `infoHandler` or `onInfo` or something else.
* I am not sure if there should be some more general scheme for overriding the various request types handled by `receiveRequest`.

james-criscuolo:
Hi @markandrus,
  I prefer `onInfo` to `infoHandler`, due to the existence of the `mediaHandler` (not a listener). Most of the functions we use for listeners use the `onX` pattern, so I'd say that makes more sense (between those two choices, anyway). The glaring difference from what I stated is the `eventHandlers` object in `hold`, which isn't set up very well to begin with anyway and needs some attention to be fixed properly.

  As for a general scheme, it seems all of the other cases already have an emission of some sort, so I think adding this is OK. More general schemes are preferable, and if there was such a scheme for this it'd probably deserve it's own PR, but I don't think that means this one would be rejected.

  Lastly, before we accept this, I hope @alexandrkozlov could test this to make sure it does indeed fix the issue. I believe it does, but confirmation would be great.

Thanks,
James

markandrus:
Thanks for the feedback, @james-criscuolo. I've renamed to `onInfo` and added a test.

alexandrkozlov:
Hi,

Sorry, I'm new on github, and I do not understand how to check out your changes. I do not see these changes nor in master nor in release branches. Could you help me?

josephfrazier:
Hi @alexandrkozlov,

If you look towards the top of this page, you'll see a link that reads [`markandrus:infoHandler`](https://github.com/markandrus/SIP.js/tree/infoHandler). This means that the changes @markandrus is proposing are on the `infoHandler` branch of his fork, which can be cloned from https://github.com/markandrus/SIP.js

Hope this helps,
Joseph

alexandrkozlov:
Hi,

All looks good for me. Please check in your changes to master.

james-criscuolo:
This PR is now  good to merge, I'll inform the other authors to do any last checks before we merge it sometime next week.

Thanks,
James

james-criscuolo:
Should've merged this a while ago, merging now (by now, I mean 6 minutes ago). Thanks again!

PR #331, opened 2016-06-02
Don't pass RTCConstraints to new RTCPeerConnections.

wpp:
This addresses github.com/onsip/SIP.js/issues/328.
Like @james-criscuolo mentioned passing the constraints around is
causing `Malformed constraints object` errors.

I've tested the changes with the following constraints:
```
var options = {
  RTCConstraints: {
    offerToReceiveAudio: true,
    offerToReceiveVideo: true
  },
  media: {
    stream: stream,
    render: {
      remote: document.getElementById('...'),
      local:  document.getElementById('...')
    }
  },
  extraHeaders: [
    'X-Browser: ' + adapter.browserDetails.browser
  ]
};
```

- Firefox 38 (Windows)
- Firefox 46.0.1(Windows + OS X)
- Chrome Version 51.0.2704.79 (Win + OS X)

Cheers

### DIFF
--- a/src/WebRTC/MediaHandler.js
+++ b/src/WebRTC/MediaHandler.js
@@ -28,7 +28,7 @@ var MediaHandler = function(session, options) {
   var servers = this.prepareIceServers(options.stunServers, options.turnServers);
   this.RTCConstraints = options.RTCConstraints || {};
 
-  this.initPeerConnection(servers, this.RTCConstraints);
+  this.initPeerConnection(servers);
 
   function selfEmit(mh, event) {
     if (mh.mediaStreamManager.on) {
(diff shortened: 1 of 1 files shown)


james-criscuolo:
Hi @wpp,
  Thanks for doing the testing on this one. It looks good to me and @alexandrkozlov also tested, so I'll ping the other authors and let this sit a few days before merging.

For some color from the  issue, it seems both [Chrome](https://code.google.com/p/webrtc/issues/detail?id=3282) and [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1033833) updated to follow [the spec](https://w3c.github.io/webrtc-pc/#rtcpeerconnection-interface) by mid-2015. Since the options we passed in this argument are always passed to createOffer/Answer, it seems that we aren't losing any functionality dropping this. Firefox flat-out ignores the 2nd argument, but Chrome will throw an error if it's 'malformed,' which is why this change was necessary.

Thanks,
James

wpp:
Awesome! Thanks @alexandrkozlov and @james-criscuolo. We are doing some more testing today, if I notice anything I'll update ASAP.

james-criscuolo:
Thanks again, @wpp!

PR #333, opened 2016-06-09
Method is `sendRequest`, not `request`

markandrus:
Hi SIP.js Team,

I found a small bug in the InviteServerContext's `terminate` method. We were calling `request` instead of `sendRequest`.


james-criscuolo:
Looks good, thanks!

Issue #75, opened 2014-07-22
InviteServerContext calls .setDescription() after session.accept()


markandrus:
I have some code that I'm testing in Node. After setting up the UA (SIP.js 0.6.1), I do

``` js
ua.on('invite', function(session) {
  session.reject();
});
```

And then I send an INVITE to my UA. But when my UA receives the INVITE, SIP.js logs

```
Tue Jul 22 2014 14:14:55 GMT-0700 (PDT) | sip.ua | INVITE received but WebRTC is not supported
```

And then I get the following error:

```
Uncaught TypeError: Cannot call method 'receiveResponse' of null
 at Object.IncomingRequest.reply (sip.js/dist/sip.js:4083:27)
 at Object.UA.receiveRequest (sip.js/dist/sip.js:8842:19)
 at Object.Transport.onMessage (sip.js/dist/sip.js:8106:17)
 ...
```

`request.reply(488)` in `UA.prototype.receiveRequest` fails because the `server_transaction` on the `IncomingRequest` is `null`. But really, I'd like to be able to `reject()` the INVITE before SIP.js sends the 488.

markandrus:
Just a followup on this issue: I had resigned myself to this behavior, until I started looking at another feature—the ability to override PeerConnection `iceServers` per-Session.

Because `inviteServerContext.accept` calls `mediaHandler.updateIceServers` with the optional STUN and TURN server arguments after the MediaHandler has been constructed and after `mediaHandler.setDescription` has been called, I needed to defer the actual construction of the PeerConnection and the call to `peerConnection.setRemoteDescription` in the MediaHandler until `inviteServerContext.accept` calls `mediaHandler.getDescription`.

So a custom MediaHandler allowed me to get the behavior I originally described in this ticket, as well as the added benefit of being able to override the PeerConnection `iceServers` per-Session. I would be happy closing this ticket. Thanks!

egreenmachine:
We believe that this issue may be resolved in the latest tagged version [0.8.0](https://github.com/onsip/SIP.js/releases/tag/0.8.0). Can you please retest with that version and let us know if this is still an issue. We will close this issue on 11/1/17 if we do not hear back. Thanks.

egreenmachine:
As stated in previous post, closing due to inactivity and proposed solution in 0.8.X. If this turns out to still be an issue, please re-open.

PR #336, opened 2016-06-23
Add hackCleanJitsiSdpImageattr UA option

josephfrazier:
Jitsi likes to send INVITEs with invalid `a=imageattr` lines like:

    a=imageattr:96 send * recv [x=[0-1920],y=[0-1080]]

When applying the SDP to a PeerConnection, Firefox doesn't tolerate
these malformed lines (here's the RFC [grammar]), and throws an error like:

    DOMException [InvalidSessionDescriptionError:
    "Failed to parse SDP: SDP Parse Error on line 21: Value too small at column 20"
    code: 0
    nsresult: 0x0]

Normally, this sort of SDP mangling would be done in a 'setDescription'
listener on the Session's MediaHandler, but SIP.js attempts to apply the
INVITE's SDP to the PeerConnection before the application is aware of the
session, as described in https://github.com/onsip/SIP.js/issues/75#issuecomment-55045403,
so we need another way to handle this situation.

This commit adds a new opt-in UA configuration option, `hackCleanJitsiSdpImageattr`,
that causes the WebRTC MediaHandler to replace invalid SDP like the
above with the following, which is valid and can be parsed by Firefox:

    a=imageattr:96 send * recv [x=[1:1920],y=[1:1080]]

[grammar]: https://tools.ietf.org/html/rfc6236#section-3.1.1

### DIFF
--- a/src/UA.js
+++ b/src/UA.js
@@ -904,6 +904,7 @@ UA.prototype.loadConfig = function(configuration) {
       hackIpInContact: false,
       hackWssInTransport: false,
       hackAllowUnregisteredOptionTags: false,
+      hackCleanJitsiSdpImageattr: false,
 
       contactTransport: 'ws',
       forceRport: false,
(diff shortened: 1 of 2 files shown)


josephfrazier:
I've reported the Jitsi bug to users@jitsi.org, per the [CONTRIBUTING guide](https://github.com/jitsi/jitsi/blob/005efc59c547adc56d0f8a6f192dc3b796407807/CONTRIBUTING.md#reporting-issues). Once the email has been approved and is available in the [list archives](http://lists.jitsi.org/pipermail/users/), I'll post the link here.

PR #352, opened 2016-09-12
Handle INFO requests in early dialogs

markandrus:
Session should handle INFO requests whenever a dialog is established, even if the dialog is early. Before this change, INFO requests were only handled when the Session's status was `STATUS_WAITING_FOR_ACK` or `STATUS_CONFIRMED`, i.e. after sending a 200 OK or receiving an ACK. Now, Session will handle INFO requests when its status is any of the following
- `STATUS_1XX_RECEIVED`: the UAC received a 101-199 response
- `STATUS_WAITING_FOR_PRACK`: the UAS sent a 18x response (and is awaiting a PRACK)
- `STATUS_WAITING_FOR_ACK`: the UAS sent a 200 OK
- `STATUS_ANSWERED_WAITING_FOR_PRACK`: the UAS sent a 18x response (and `accept` was called before receiving a PRACK)
- `STATUS_EARLY_MEDIA`: the UAC received a 18x response or the UAS received a PRACK
- `STATUS_CONFIRMED`: the UAS received an ACK

Per [RFC 6086, Section 4.2.1, "INFO Request Sender"](https://tools.ietf.org/html/rfc6086#section-4.2.1):

> The construction of the INFO request is the same as any other
> non-target refresh request within an existing invite dialog usage as
> described in Section 12.2 of RFC 3261.

And [RFC 3261, Section 12.1, "Creation of a Dialog"](https://tools.ietf.org/html/rfc3261#section-12.1):

> Dialogs are created through the generation of non-failure responses
> to requests with specific methods.  Within this specification, only
> 2xx and 101-199 responses with a To tag, where the request was
> INVITE, will establish a dialog.  A dialog established by a non-final
> response to a request is in the "early" state and it is called an
> early dialog.

### DIFF
--- a/src/Session.js
+++ b/src/Session.js
@@ -674,7 +674,8 @@ Session.prototype = {
             this.status === C.STATUS_WAITING_FOR_ACK ||
             this.status === C.STATUS_ANSWERED_WAITING_FOR_PRACK ||
             this.status === C.STATUS_EARLY_MEDIA ||
-            this.status === C.STATUS_CONFIRMED) {
+            this.status === C.STATUS_CONFIRMED ||
+            this.dialog) {
           if (this.onInfo) {
             return this.onInfo(request);
           }
--- a/src/Session.js
+++ b/src/Session.js
@@ -669,7 +669,12 @@ Session.prototype = {
         }
         break;
       case SIP.C.INFO:
-        if(this.status === C.STATUS_CONFIRMED || this.status === C.STATUS_WAITING_FOR_ACK) {
+        if (this.status === C.STATUS_1XX_RECEIVED ||
+            this.status === C.STATUS_WAITING_FOR_PRACK ||
+            this.status === C.STATUS_WAITING_FOR_ACK ||
+            this.status === C.STATUS_ANSWERED_WAITING_FOR_PRACK ||
+            this.status === C.STATUS_EARLY_MEDIA ||
+            this.status === C.STATUS_CONFIRMED) {
           if (this.onInfo) {
             return this.onInfo(request);
           }


markandrus:
Hey SIP.js team,

I cut this first commit thinking it would be sufficient to catch all the early dialog states, but after some more manual testing, I see that there is a scenario where, if the UAC's INVITE is immediately responded to by a 200 OK, we will create a confirmed Dialog, but not update the Session's `status` until the MediaHandler returns in the next tick. So there is a short window where INFO requests will be dropped. It seems like one way to fix this is to check the `dialog` property (implemented in the second commit), but I am wondering,

* Should this in-between state should be represented as a `status` on the Session?
* Do we need to check `earlyDialogs` property as well?

Thanks,
Mark

james-criscuolo:
Hi @markandrus,
  Thanks for the PR! My memory is a bit hazy on `earlyDialogs`, as that code is now nearly 3 years old I believe, but `earlyDialogs` are mainly for branching scenarios with early media. We need to be able to negotiate multiple sets of media, then throw all but one away once one is confirmed. So in terms of the RFC, it sounds like yes, you would have to check `earlyDialogs`, as you only get one of those after a 1xx is received (and on top of that it must be reliable), but before a 2xx received.

As for the in-between state, is there a reason we can't update the status before going to the mediaHandler? And if not, is there a valid name for this period of time? Most, if not all, of our states involve something with SIP, and this gap is for setting up media with the browser, so naming it would seem hacky. I'd prefer to update the state before going to the mediaHandler, but there may be good reasons not to do that.

Thanks,
James

egreenmachine:
We are in the process of updating SIP.js to typescript. This will not merge cleanly once that is complete. If you would like to get this in before then please address the issues outlined and we will merge it before we do the conversion. Otherwise this will be closed.

markandrus:
Please proceed with the conversion!

PR #360, opened 2016-10-31
Added a hack to strip TCP candidates from the sdp

mrothstein:
This hack enables the removing of candidates of type TCP from the SDP provided in the proxy server ACK. To turn on the hack, set session.ua.configuration.hackStripTcp to "true."


kirill-konshin:
The team has verified that the original issue is gone after the fix. Please go ahead and merge it.

PR #365, opened 2016-12-18
Basic documentation changes and repository setup

rts-rob:
These are the basic changes to get us going while maintaining proper credit to the SIP.js and JsSIP folks.


rts-rob:
Sorry - wrong direction.

PR #372, opened 2017-01-13
fix bug in subscription refresh

davehorton:
if first time around we requested (say) 300 and got 299, then during refresh we requested 300 and got 300.  this was incorrectly triggering the "Expires header in a 200-class response to SUBSCRIBE with a higher value than the one in the request" warning, following by an unsubscribe

### DIFF
--- a/src/Subscription.js
+++ b/src/Subscription.js
@@ -89,7 +89,7 @@ SIP.Subscription.prototype = {
   },
 
   receiveResponse: function(response) {
-    var expires, sub = this,
+    var expires, requestedExpires, sub = this,
         cause = SIP.Utils.getReasonPhrase(response.status_code);
 
     if ((this.state === 'notify_wait' && response.status_code >= 300) ||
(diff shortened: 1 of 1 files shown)


james-criscuolo:
Hi @davehorton,
  Instead of pulling the `requestedExpires` out of the extraHeaders, I'd rather just set that on `this` during creation (around line 31). This is a passed in value anyway, so we can skip pulling apart `extraHeaders`. You could likely only set `this.requestedExpires` there and set `this.expires` to the negotiated value later, but that would require a few more changes (not a bad thing, just need to be more careful). Besides that, I believe this PR would be clean.

james-criscuolo:
Just put this simplified change (discussed above) on master, thanks again!

davehorton:
Thanks!
On Sep 28, 2017, at 1:57 PM, James Criscuolo <notifications@github.com> wrote:

Just put this simplified change (discussed above) on master, thanks again!

PR #373, opened 2017-01-16
Fix for missing ACK on PRACK on incoming call (incoming calls were terminating after 30 seconds if rel100 were used)

roginvs:
If sipjs were used with rel100, then on incoming calls sipjs were not answering with ACK on PRACK (which is required by RFC). This caused asterisk to terminate call after ~30 seconds due to missing ACK.

### DIFF
--- a/src/Session.js
+++ b/src/Session.js
@@ -1588,7 +1588,7 @@ InviteServerContext.prototype = {
           //REVISIT
           this.mute();
         }
-      } else if(this.status === C.STATUS_EARLY_MEDIA) {
+      } else if(this.status === C.STATUS_EARLY_MEDIA || this.status === C.STATUS_WAITING_FOR_ANSWER) {
         request.reply(200);
       }
       break;


james-criscuolo:
Hi @roginvs,
  Can you provide a gist of the broken signaling? The RFC is very explicit here, you are right, but I'm more curious how you are ending up in this case. If you don't set the rel100 value on your UA, or messages come out of order in a potentially troublesome way, this may be possible, but if SIP.js is expecting a PRACK, then your PR should not be necessary. In other words, I believe the issue may be deeper than your fix, or a potential configuration error.

Thanks,
James

roginvs:
[chrome-log.txt](https://github.com/onsip/SIP.js/files/717089/chrome-log.txt)
This is chrome log. As we can see, on original version sipjs ignores PRACK and continues to resend same Session Progress.

james-criscuolo:
The problem stems from: https://github.com/onsip/SIP.js/blob/master/src/Session.js#L1035

I think the reasoning was that we didn't want to force early negotiation, but if you were to switch your 100rel to supported instead of required you would be forcing early negotiation anyway. I think the conditional needs to go away, and the status set to `WAITING_FOR_ANSWER` only done if 100rel doesn't occur (this should probably move to the conditional that checks if `do100rel` should be called).

I dont think this will cause further issues, but this is the more correct way to fix this problem, and will also do the 183 for you (assuming you were manually making that call happen).

roginvs:
We are using 100rel because without REQUIRED rel100 sipjs does not play media at early media stage on outgoing calls (this is quite a issue for us).
I had not understand how exactly do you want me, should I close this pull request?

james-criscuolo:
There is a clear issue, I'm just saying the solution should be handled differently. Whenever we receive a PRACK, we should be in `WAITING_FOR_PRACK`, so instead of allowing the OK to be sent in more situations (such as `WAITING_FOR_ANSWER`, I'd rather get your situation to be in `WAITING_FOR_PRACK` state. My explanation above outlined how that would be doable.

roginvs:
Could it be (in theory) that we will receive OK before PRACK, in `WAITING_FOR_PRACK` state? I suppose on UDP this have non-zero probability, for example it wss server will be a some sort of webrtc->udp sip proxy.

james-criscuolo:
Closing due to inactivity and need for rework due to typescript update.

PR #375, opened 2017-01-18
Allow RTCP Multiplexing to be negotiated via configuration parameter

### LINKED ISSUE #374: Multiplexing RTCP required in Chrome Canary
Per [this announcement](https://groups.google.com/forum/#!topic/discuss-webrtc/eM57DEy89MY):

> As of the most recent Chrome Canary build, the default RTCP multiplexing policy is "require", instead of "negotiate". This will affect the next Chrome release, M57.

> For any application that doesn't yet support RTCP multiplexing, you can get the old behavior by explicitly setting the RTCRtpMuxPolicy to "negotiate" in the RTCConfiguration.

Some endpoints (Asterisk, for example) do not yet support rtcp-mux and will therefore stop working after the next Chrome release.

seanbright:
Fixes #374

### DIFF
--- a/src/UA.js
+++ b/src/UA.js
@@ -1175,6 +1175,7 @@ UA.configuration_skeleton = (function() {
       "replaces",
       "userAgentString", //SIP.C.USER_AGENT
       "autostart",
+      "rtcpMuxPolicy",
       "stunServers",
       "traceSip",
       "turnServers",
@@ -1573,6 +1574,12 @@ UA.configuration_check = {
         }
       }
       return turnServers;
+    },
+
+    rtcpMuxPolicy: function(rtcpMuxPolicy) {
+      if (typeof rtcpMuxPolicy === 'string') {
+        return rtcpMuxPolicy;
+      }
     },
 
     userAgentString: function(userAgentString) {
--- a/src/WebRTC/MediaHandler.js
+++ b/src/WebRTC/MediaHandler.js
@@ -418,7 +418,15 @@ MediaHandler.prototype = Object.create(SIP.MediaHandler.prototype, {
       this.peerConnection.close();
     }
 
-    this.peerConnection = new SIP.WebRTC.RTCPeerConnection({'iceServers': servers});
+    var connConfig = {
+      iceServers: servers
+    };
+
+    if (config.rtcpMuxPolicy) {
+      connConfig.rtcpMuxPolicy = config.rtcpMuxPolicy;
+    }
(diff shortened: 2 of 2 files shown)


seanbright:
I'm not sure if this is a desirable way to implement this, so feel free to toss it if there is a more idiomatic solution.

james-criscuolo:
Hi @seanbright,
  This PR looks good to me. I'll give another day or 2 (or until Monday, I guess) for other authors/users to take a look, but I think it'll be good.

Thanks,
James

GMLudo:
Hi @seanbright,

Thanks for the pull request, it looks good to me.

When it will be merged, I vote for a new release of sip.js: We have some integrators that use your library, easier for them to accept an official release than a patch, ideally.

Have a nice week-end.

egreenmachine:
@seanbright thanks for the initial pr, we really appreciate the community support. 

Here are my thoughts on this, and I am not particularly sold one way or the other so if people disagree we can certainly go another route. 

I think that this is not necessarily a "hack" in our definition of the word and should not be named as such. We should call it rtcpMuxPolicy on the UA. 

If we are going to make this configurable from the ua it should pass the value as is to the peerconnection. So we shouldn't do a Boolean and then fill in a string. The user should be responsible for filling in the string. This will all obviously be documented. 

Lastly, and here is where I am really unsure about what to do. I think it would be best at this time to default the value to negotiate. The upside to this is users can upgrade and everything will continue working as is. The downside is eventually the browsers may not allow this at all in a future update and then we will have the change the default behavior of the library. 

Obviously once this is merged we can rush a release so that users can test in the browsers. 

Let me know your thoughts. Thanks!

P.S. Bonus points for unit tests ;)

seanbright:
@egreenmachine I've updated the PR based on your comments. What specifically should an acceptable unit test be testing here?

egreenmachine:
Hey @seanbright, sorry for the slow response on this. The changes look good, @james-criscuolo and I were having some discussion internally about what we want to see here and researching what the browsers are doing. 

Based on our research and discussion I think we want to not default to "negotiate". The discussion on the Chromium boards essentially say that it is "high risk" that this option will be deprecated, so we do not want to default to something that we will have to remove. So if you make the change to remove the default, we will take this then do a release. Sorry for flip flopping on it so much.

Thanks!

seanbright:
@egreenmachine, I've updated. Let me know if additional changes are required.

PR #385, opened 2017-02-10
Removed references to closed WebSocket

matthieusieben:
Whenever the WebSocket is closed, remove any reference to it.

Also, remove the listeners setup on the WebSocket in case some other part of the code does nasty stuff an keeps a reference to it. In that case, only the WebSocket will remain in memory and not the whole UA instance.


james-criscuolo:
Hi @matthieusieben,
  Thanks for the PR on this one! It looks good, I'll let it sit for the weekend in case another author/person spots something, then I'll merge early next week.

Thanks again,
James

james-criscuolo:
Just merged, thanks again!

PR #397, opened 2017-03-22
Added an options argument to session.hold()

chariot6:
The unhold method has an options argument already. Both hold() and unhold() pass these options on to sendReinvite().


james-criscuolo:
Sorry for the delay, but this looks good, I'll merge. Thanks!

Issue #401, opened 2017-04-20
Configurable UA.stop() on page unload


VicNumber21:
I'd like to ask to make UA.stop() on 'unload' event to be configurable.
I think, most of users need the current behavior but I need UA.stop() not to be called to be able to recover call after page refresh.
If you are ok with making it configurable, I can implement it myself and create pull request for you.

To be more clear, I'm about the following piece of code in UA.js
```
  if (typeof environment.addEventListener === 'function') {
    // Google Chrome Packaged Apps don't allow 'unload' listeners:
    // unload is not available in packaged apps
    if (!(global.chrome && global.chrome.app && global.chrome.app.runtime)) {
      this.environListener = this.stop.bind(this);
      environment.addEventListener('unload', this.environListener);
    }
  }
```

I'd like to have it like the following:
```
  if (this.configuration.autostop && typeof environment.addEventListener === 'function') {
    // Google Chrome Packaged Apps don't allow 'unload' listeners:
    // unload is not available in packaged apps
    if (!(global.chrome && global.chrome.app && global.chrome.app.runtime)) {
      this.environListener = this.stop.bind(this);
      environment.addEventListener('unload', this.environListener);
    }
  }
```

Default value of 'autostop' could be true.

egreenmachine:
We would definitely take a pull request for this. I would like to maintain backwards compatibility, so the default would be to autostop. Thanks!

VicNumber21:
Please look at this change.

VicNumber21:
Solved now, thanks a lot!

PR #403, opened 2017-04-21
Add autostop setting to allow to disable UA.stop on unload event

VicNumber21:
See #401


james-criscuolo:
Looks good, I'll merge, thanks!

Issue #390, opened 2017-02-17
Release an NPM browser build


sdemjanenko:
The bower.json file references a `dist` directory but the NPM version does not.  The `var pkg = require('../package.json')` makes it necessary for me to use a `json-loader` in my webpack config to get this library to work.  Your `npm publish` step should take care of properly building a `dist` that doesn't require any special loaders on my end.

egreenmachine:
We agree that we should probably do this. We will look into doing this for our next tagged release.

subramaniashiva:
I am also facing issues without a dist version in npm. Please include in the upcoming release. Thanks for spending your time on this feature request :)

brunowego:
Yes, please add a production ready dist.

james-criscuolo:
This is now released with 0.7.8 so I'm going to close this issue, please let us know if there are any further issues.

PR #420, opened 2017-05-27
Publish dist files

brunowego:
I am not familiar with this repo and I not know if this is the correct way. Please let me know if need add more validations for grunt.

#390


james-criscuolo:
Hi @brunowego,
  We have a release branch where we include the dist files and publish from. We plan to do a release this week, at which point we plan to address #390.

Thanks,
James

Issue #407, opened 2017-04-27
accept() will override progress()'s mediaHints


matthieusieben:
On an incoming call (ServerContext) if you perform a `progress({ media: { render: { ... } } })`, the media option will be saved in `this.mediaHints`. If you then perform an `accept(null)`, the `session.mediaHints` will be overwritten with the configuration default, meaning that the mediaHints do not represent the actual hints requested by the user.

There may even be a race condition where the session is accepted after the call to progress, but before the async resolve handler of `mediaHandler.setDescritpion()`. In that case, the render media provided through the progress options will not be used at all.

matthieusieben:
My current hack is as follows:
```js
session.accept =  function accept (options) {
  if (this.mediaHint && !isMediaElement(options)) {
    options = options || {}
    options.media = options.media || this.mediaHint
  }
  return InviteServerContext.prototype.accept.call(this, options)
}
```

egreenmachine:
We believe that this issue may be resolved in the latest tagged version [0.8.0](https://github.com/onsip/SIP.js/releases/tag/0.8.0). Can you please retest with that version and let us know if this is still an issue. We will close this issue on 11/1/17 if we do not hear back. Thanks.

matthieusieben:
Sadly, I no longer work for the same company and will not be able to check.

You should however be able to simulate the behaviour described here by reasoning on your new code.

Closing issues because you assume that your refactor fixed them all is a mistake that may affect the quality of your product in my opinion.

egreenmachine:
The code was not simply refactored, but the media handling was completely re-written. I have tested this case pretty extensively, and this was simply a request for you to test your application to make sure I did not miss your specific case.

james-criscuolo:
For further reference, all of the issues that were given the above ('please test or we will close') were in a similar boat, where  they were specifically posted on because they were directly addressed by the changes and tested. As @egreenmachine, this is more of a confirmation. We did not put this message on all of our issues, just the ones directly affected/tested.

matthieusieben:
Alright, sorry if my comment was a little rude ;-)

Issue #406, opened 2017-04-26
Do not render audio before call is accepted


matthieusieben:
When receiving an incoming call, the incoming stream should not be attached to the remote renderer before the call is accepted. This is particularly true when early media is established.

matthieusieben:
My current workaround (only works if #407 is used too)

```js

session.progress = function (options) {
  options = Object.create(Session.desugar(options))

  Utils.optionsOverride(options, 'media', 'mediaConstraints', true, this.logger, this.ua.configuration.media)
  this.mediaHint = options.media

  // @hack: https://github.com/onsip/SIP.js/issues/406
  //
  // In case of success of the getDescription() method, the mediaHandler will
  // call `self.render()`. This means thar as soon as early media is
  // established, the remote media will start rendering on incoming sessions.
  // If the remote phone sends early media before the call was accepted, then
  // the callee will be able to hear the caller before he answered the call.
  // In order to prevent this, lets save the remote renderer, and set it back
  // later on, when the call is accepted.
  if (this.status !== Session.C.STATUS_CONFIRMED) {
    const remote = this.mediaHint.render && this.mediaHint.render.remote
    if (remote) {
      this.mediaHint.render.remote = null
      this.on('accepted', function () {
        if (!this.mediaHint.render.remote) {
          this.mediaHint.render.remote = remote
          this.mediaHandler.render()
        }
      })
    }
  }

  return InviteServerContext.prototype.progress.call(this, options)
}
```

matthieusieben:
Note that this was an issue with Polycom phones sending voice instead of ringback through early media.

egreenmachine:
Between this and #407 it seems like you want to do early media, but not render it until the call is accepted, correct?

matthieusieben:
Indeed. And conceptually, it does seem strange to connect the renderer and the incoming stream before one accepted the call, doesn't it?

egreenmachine:
Not necessarily. For calls to the PSTN for example, you may want to hear the ringback or early media. There are instances where error messages about lines being disconnected will not ever accept the call and will be done only via early media.

I think what you want to do is possible here. I have to review the media handling code.

matthieusieben:
Note that i do wish to ear incoming media before they are accepted for *outgoing* calls, not for *incoming* calls.

egreenmachine:
[...]

Issue #393, opened 2017-03-06
Hold/Unhold is not working


anshkatriya:
When caller or callee hold current call, it works pretty fine. But when caller or callee unhold that call, at that time not able to get opponent voice. It seems like it's issue of attaching remote stream again to video element.

egreenmachine:
Can you provide some logs of this? It is most likely an issue with the underlying browser, but if there is something we can do to get it going we would be happy to add that to the library.

anshkatriya:
Here, I am adding SIPjs logs with SIP trace.

sip-0.7.3.js:2884 Wed Apr 26 2017 18:51:58 GMT+0530 (IST) | sip.ua | configuration parameters after validation:
sip-0.7.3.js:2884 Wed Apr 26 2017 18:51:58 GMT+0530 (IST) | sip.ua | Â· viaHost: "ntq512jm7hh2.invalid"
sip-0.7.3.js:2884 Wed Apr 26 2017 18:51:58 GMT+0530 (IST) | sip.ua | Â· uri: sip:1000107@demo.example.com
sip-0.7.3.js:2884 Wed Apr 26 2017 18:51:58 GMT+0530 (IST) | sip.ua | Â· wsServers: [{"ws_uri":"wss://demo.example.com:7443","sip_uri":"<sip:demo.example.com:7443;transport=ws;lr>","weight":0,"status":0,"scheme":"WSS"}]
sip-0.7.3.js:2884 Wed Apr 26 2017 18:51:58 GMT+0530 (IST) | sip.ua | Â· password: NOT SHOWN
sip-0.7.3.js:2884 Wed Apr 26 2017 18:51:58 GMT+0530 (IST) | sip.ua | Â· registerExpires: 600
sip-0.7.3.js:2884 Wed Apr 26 2017 18:51:58 GMT+0530 (IST) | sip.ua | Â· register: false
sip-0.7.3.js:2884 Wed Apr 26 2017 18:51:58 GMT+0530 (IST) | sip.ua | Â· registrarServer: sip:demo.example.com
sip-0.7.3.js:2884 Wed Apr 26 2017 18:51:58 GMT+0530 (IST) | sip.ua | Â· wsServerMaxReconnection: 3
sip-0.7.3.js:2884 Wed Apr 26 2017 18:51:58 GMT+0530 (IST) | sip.ua | Â· wsServerReconnectionTimeout: 4
[...]

Issue #349, opened 2016-09-02
Chrome/Opera throw "DOMException: The play() request was interrupted by a new load request." on session.accept()


avillacis:
I have this webphone project that uses SIP.js 0.7.5 . The javascript code is supposed to be a straightforward implementation of a webphone GUI. My HTML templates output a pair of <video> tags that are statically picked up when generating or accepting calls.

My available browsers are:
firefox-48.0.1-1.fc24.x86_64
google-chrome-stable-53.0.2785.92-1.x86_64
opera-developer-41.0.2329.0-0.x86_64

The webphone is supposed to connect to an Asterisk 11.21.0 instance.

Under Firefox, both call origination and call answering work correctly from the webphone.

Under Opera or Google Chrome, call origination works correctly. However, call answering fails. There is no audio, and the console shows this error:

sip-0.7.5.js:11567 Uncaught (in promise) DOMException: The play() request was interrupted by a new load request.
    attachMediaStream @ sip-0.7.5.js:11567
    attachAndPlay @ sip-0.7.5.js:11594
    render @ sip-0.7.5.js:11601
    (anonymous function) @ sip-0.7.5.js:11283
    render @ sip-0.7.5.js:11280
    sdpCreationSucceeded @ sip-0.7.5.js:6041

(I was originally using the minified version, but switched to the development version in order to have a backtrace)

I am well aware of issue #315. Please, do NOT close this as a duplicate. The referenced issue never got an answer about what to do to work around the exception. I am not calling a play() method anywhere in my code, and I am not touching or referencing the MediaStreams anywhere, since I have no need for it.

Following what I understand from the comment on #315 by @svarunan , I tried dynamically adding the <video> tags right before accepting the call, but this had no effect on the bug. I have also tried calling .removeAttribute('src') on the <video> tag, following a suggestion from a comment on a similar package, also without success.

avillacis:
Just to be clear, I also implemented the suggestion to .pause() then remove the tag by @svarunan , also without success.

avillacis:
Using the development SIP.js, I found out that inserting:

`if (!element.paused) element.pause();`

right before the call to environment.revokeObjectURL() (of which there is only one instance in the nested function attachMediaStream()) will silence the DOMException under Chrome. However, it does *not* fix the absence of audio.

DanEdwards61:
Chrome has an issue open for this. Not sure if this is your case or not, but see https://bugs.chromium.org/p/chromium/issues/detail?id=593273

[...]

PR #426, opened 2017-06-23
Session description refactor

egreenmachine:
# Session Description Refactor
This is a pretty big code refactor in an attempt to simplify SIP.js and the handling of SDP and media. The idea is that SIP.js should be more strictly a signaling library, and not get involved in the handling of the actual media. Handling the media will be the responsibility of the end application. That being said, some of the simplicity of getting up and running with SIP.js has been removed. I am hoping to write a cover library that really makes getting up and running very simple, and then having users who want more control come and use SIP.js directly. With all that being said, let me outline the changes.

## hacks.js
`hacks.js` and all the functions within have been removed. These all seemed to be related to old browser problems and SDP manipulation. The idea was that we could use it as a reference to see which browsers were working as closely to intended behavior as possible. The reality of it was, that once something was added, it was almost never removed. Some hacks never made it into this file. And it was not signaling related, but SDP in almost all cases. There is a new way to do SDP manipulation which will be outlined later.

## environment.js
The deprecation of `environment.js`. The idea with the environment was to essentially act a shim. It had a list of things that any environment running SIP.js had to implement, and then SIP.js would work as expected. The reality is that it caused more trouble than it solved. It required some things that only the browsers implemented and were not actually required for the library to function, like functions required by the media handler. I have started the process of removing everything from here and the hope moving forward is to offer up a different library that acts a better shim in the future.

## Session Description Handler
The change of `MediaHandler` to `SessionDescriptionHandler`. This is the big one, and it starts with the name. The name MediaHandler does not accurately describe what is actually being handled. SIP.js does not want to handle media directly, that is what WebRTC is for. We want to handle the signaling. The new `SessionDescriptionHandler` will attempt to do just that, it will set up the PeerConnection and signal that is has done so. The application will then be responsible for doing what it wants with the media, similar to how the application is responsible for implementing it's own user interface. 
The mute and unmute options are no longer available on a session.
### Example of media handling in application
```
session.on('accepted', function() {
    var streams = session.sessionDescriptionHandler.peerConnection.getRemoteStreams();
    var remoteElem = document.getElementById('remoteAudio');
    remoteElem.srcObject = streams[0];
    remoteElem.play();
});
```
### SDP Modifiers
The next thing that has changed with the Session Description Handler is the mangling of SDP. We have come up with a scalable way to modify SDP before passing it to the PeerConnection and after receiving it from the PeerConnection. These are SDP modifiers. An SDP modifier function is simply a function that takes SDP and returns a promise which resolves with the modified SDP. The session description handler then runs through this array in order passing the result of the previous manipulation to the next function. Currently getting these modifiers into the session description handler is a bit fickle, but I hope to make this process easier over time.
### Hold
Hold is now done by just defining a `holdModifier` on the session description handler. When the hold function on the session is called, the session simply calls `getDescription` from the handler and passes it the `holdModifier` to mangle the sdp appropriately. 
### Renegotiation
Renegotiation now works correctly. This is tested in Chrome (59+) and Firefox (54+). There is no longer any fake hold being done with the media streams.
### Creating your own handler
Just implement the functions in the outline below, wrap it in a factory. That is it. No need to deal with rendering, or muting, etc. That is for the application to worry about.
```
SessionDescriptionHandler.prototype = Object.create(EventEmitter.prototype, {

  close: {value: function close () {}},

  /**
   * @param {Object} [mediaHint] A custom object describing the media to be used during this session.
   */
  getDescription: {value: function getDescription (constraints, modifier) {
    // keep jshint happy
    constraints = constraints;
    modifier = modifier;
  }},

  /**
   * Check if a SIP message contains a session description.
   * @param {SIP.SIPMessage} message
   * @returns {boolean}
   */
  hasSessionDescription: {value: function hasSessionDescription (contentType) {
    // keep jshint happy
    contentType = contentType;
  }},

  /**
   * Set the session description contained in a SIP message.
   * @param {SIP.SIPMessage} message
   * @returns {Promise}
   */
  setDescription: {value: function setDescription (sessionDescription, constraints, modifier) {
    // keep jshint happy
    sessionDescription = sessionDescription;
    constraints = constraints;
    modifier = modifier;
  }},

  holdModifier: {value: function holdModifier(sdp) {
    return SIP.Utils.Promise.resolve(sdp);
 }}
});
```
### Session Description Handler Options
These options should now all be contained within a single key on the UA configuration. Validation is still done on UA side, but it should change in the future to be done on the SessionDescriptionHandler itself.

## Removing WebRTC Dependencies
All WebRTC dependencies should be included in the default session description handler provided by SIP.js. The goal here is to allow SIP.js to run in environments that do not have WebRTC by simply overriding the default `sessionDescriptionHandlerFactory`. This can be done even with compiled versions of SIP.js by just passing the new factory as a UA parameter. There are no longer any mysteries of `isReady()` or needing to call WebRTC functions to get SIP.js to run.

## TODO's
There are a few remaining items left to do.
1. Documentation. Documentation. Documentation. Lots of new documentation needs to be written and old documentation cleaned up. 
2. Tests. A lot of unit tests weren't actually testing units, and did not apply once the old media handler stuff was removed.
3. Bug fixes and edge cases. There are definitely some things not working with this, refer is one that jumps to mind that is most likely broken.
4. Finish cleaning up the environment
5. Clean up passing of arguments to the sessionDescriptionHandler and make that consistent.
6. Clean up other issues and PR's that are resolved by this.

Lastly, I would like for this pull request to become SIP.js 0.8.0 release. I expect there are definitely some things that I have missed and some bugs in my code. I encourage the community to try it out and give me your feedback. I am sure a few more commits will make it on to this branch before we merge and release.


src/WebRTC/SessionDescriptionHandler.js
```diff
@@ -0,0 +1,414 @@
...
+  }},
+
+  /**
+   *
+   */
+  getDescription: {writable: true, value: function (constraints, modifiers) {
+    var self = this;
+
+    // We will override the saved constraints with new constraints if provided, otherwise use saved constraints
+    this.constraints = constraints || this.constraints || {};
```
egreenmachine: This could mess up the underlying WebRTC library if the `constraints` argument passed to `getDescription` is an empty object.

egreenmachine:
I have made the `SessionDescriptionHandler` a bit more flexible by changing the `constraints` passed to it via `getDescription` and `setDescription` to a property on an `options` object. The `options` could be any object you like, so that using your `SessionDescriptionHandler` can just handle it as it needs to. If you want to use `constraints` you can, but you are not required to.

On the default WebRTC `SessionDescriptionHandler` you can redefine the `PeerConnection` by passing the `peerConnectionOptions` property to the options of a `set/getDescription` call. This would be passed down all the way from the application layer functions. This is to not break #260 and #250. 

I have also separated the options passed by the `UA` to the `SessionDescriptionHandlerFactory` as it's own set of options. This is separate from the options held onto by the session to pass to `set/getDescription`. Again, this just adds some extra flexibility. 

Lastly, the `UA` will now only check to see that the options for the `SessionDescriptionHandlerFactory` is an object. It is up to the handler itself to verify these options for itself. As a note here `iceServers` and `turnServers` no longer have their own place on the UA options and should instead be defined to the spec of the `SessionDescriptionHandler`. In the default case this is the WebRTC spec for [RTCConfiguration](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/RTCPeerConnection#RTCConfiguration_dictionary).

src/RegisterContext.js
```diff
@@ -211,8 +211,7 @@ RegisterContext.prototype = {
     options = options || {};
 
     if(!this.registered && !options.all) {
-      this.logger.warn('already unregistered');
-      return;
```
egreenmachine: Hmmmm... it is not immediately clear to me how removing this `return` breaks all of our tests.

egreenmachine: It is in the test cleanup and how we are cleaning up the UA that is causing the tests to break. We are blindly calling unregister in places, and expecting it to return out of the function if it is not registered.

The issue is, that it is valid to send an unregister even if you are not registered. There are cases where you would want to do this. Either we need to re-write the tests, and confirm that we are not breaking anything. Or we need to keep the current behavior and add some extra behavior to allow this to continue in certain situations.

egreenmachine:
Just wanted to make note of the following open issues/PRs that would be affected and potentially closed once this PR is complete:
#427 #407 #406 #393 #349 #337 #326 #300 #275 #249 #242 #228 #182 #138 #75 #7 #259

src/WebRTC/SessionDescriptionHandler.js
```diff
@@ -0,0 +1,446 @@
...
+  createOfferOrAnswer: {writable: true, value: function createOfferOrAnswer () {
+    var self = this;
+    var methodName;
+    var pc = self.peerConnection;
+    // TODO: Lock?
+    // TODO: We need to lock on the setRemoteDescription so that it cannot be done 2x
+
+    methodName = self.hasOffer('remote') ? 'createAnswer' : 'createOffer';
+
+    return SIP.Utils.promisify(pc, methodName, true)()
```
egreenmachine: Need to stop using promisify as it breaks Safari

src/WebRTC/SessionDescriptionHandler.js
```diff
@@ -0,0 +1,446 @@
...
+        };
+        // TODO: This emit does not match what we are doing for getDescription... kind of re: modifier
+        self.emit('setDescription', rawDescription);
+
+        return self.peerConnection.setRemoteDescription(new self.WebRTC.RTCSessionDescription(rawDescription));
+      })
+      .catch(function modifierError(e) {
+        self.logger.error("The modifiers did not resolve successfully");
+        self.logger.error(e);
+        throw e;
```
egreenmachine: We should log the SDP that was failed to set. It would be helpful for debugging. Also if it was an offer or answer.

src/WebRTC/SessionDescriptionHandler.js
```diff
@@ -0,0 +1,446 @@
...
+        var rawDescription = {
+          type: self.hasOffer('local') ? 'answer' : 'offer',
+          sdp: sdp
+        };
+        // TODO: This emit does not match what we are doing for getDescription... kind of re: modifier
+        self.emit('setDescription', rawDescription);
+
+        return self.peerConnection.setRemoteDescription(new self.WebRTC.RTCSessionDescription(rawDescription));
+      })
+      .catch(function modifierError(e) {
```
egreenmachine: This catch may be in the wrong spot...

src/WebRTC/SessionDescriptionHandler.js
```diff
@@ -0,0 +1,446 @@
...
+        self.logger.error(e);
+        throw e;
+      })
+      .then(function setRemoteDescriptionSuccess() {
+        self.emit('setRemoteDescription', self.peerConnection.getRemoteStreams());
+        // TODO: invite w/o sdp case, receiving call case
+        self.emit('confirmed', self);
+      })
+      .catch(function setRemoteDescriptionError(e) {
+        self.emit('peerConnection-setRemoteDescriptionFailed', e);
```
egreenmachine: This failure should print the SDP that was failed to be set as well as whether it was an offer or answer.

egreenmachine:
@wpp this should be in a testable state now in Chrome/Firefox/Opera/Safari. Still have not looked into Edge yet. Let me know your thoughts.

wpp:
Hi,

took this for a spin on Chrome, FF (54.0.1) and STP. From a library users perspective here is what I had to update (_might be helpful for release notes_):

    // Change
    this.session.getRemoteStreams()
    this.session.getLocalStreams()
    // to
    this.session.sessionDescriptionHandler.peerConnection.getLocalStreams
    this.session.sessionDescriptionHandler.peerConnection.getRemoteStreams

(Short hand was pretty nice...just saying :))

We also have a listener sitting on 'getDescription' which tries to assign a modified SDP:

```
  this.session.sessionDescriptionHandler.on('getDescription', this.addOriginToSDP);
  addOriginToSDP(sdpWrapper) {
    sdpWrapper.sdp = sdpWrapper.sdp.replace('o=-', 'o=something');
  }
```

This results in an error on STP:

>TypeError: Attempted to assign to readonly property.

Last but not least: Due to the lack of VP8 support in STP I tried an audio only call.
ATM you're adding all tracks to the peer connection, which results in the video section being
added to the SDP. If the remote party can't match the video specs, we can't set it's remote Description. I.e:

```
stream.getAudioTracks().forEach(function (track) {
    self.peerConnection.addTrack(track, stream);
});
```

Other than that, I didn't notice any issues.
Thanks for working on this,
LMK if you need more help and have a great weekend.

egreenmachine:
@wpp Thanks for the feedback. Let me try and address some of these issues.

I know that the shorthand was nice, but it was not a scalable API. I may add it as an optional pass through on the session for the sessionDescriptionHandler to implement, but the internals of SIP.js will never call it. I do plan on making a "simple" wrapper that makes the simple case of making a call a lot easier, but it would still not solve the issue for power users like yourself.

So to apply modified SDP, just use the modifiers! That is what they are there for, no more nasty hacks in your application. The same could be done for removing VP8 streams. You can add as many modifiers as you want either to the constructor (passed through the factory, and options available on the UA) or when you call invite, progress, or accept.

james-criscuolo:
While this Pull Request is not a great place for the code I've pushed this morning, this branch is sort of our "0.8.0" release branch, so when people test it'd be good to get this stuff tested as well.

My changes were to replace some of our older and outdated dependencies with more current tools. Basically:

- grunt task running -> npm scripts
- browserify -> webpack
- jshint -> eslint
- grunt runs jasmine -> karma runs jasmine

I also updated our peg.js version to the latest version, with one added benefit of logging within the pegjs file if it is built to include that (one line change in webpack config). We weren't able to use webpack@latest, but I've [opened an issue](https://github.com/eploko/pegjs-loader/issues/12) to hopefully get that worked out.

egreenmachine:
Should closing the WebRTC SessionDescriptionHandler stop all of the tracks that are on it's PeerConnection? Or should this be the responsibility of the application.

egreenmachine:
Very frustrating afternoon trying to figure out why our results from a call to `createAnswer` is returning a different result in my code than the old media handler code. The issue comes down to the fact that Chrome deprecated `offerToReceiveVideo` between the callback and promise versions of `createAnswer`. The [WebRTC spec](https://w3c.github.io/webrtc-pc/#dom-rtcofferoptions-offertoreceiveaudio) states that this value is only listened to if it is a non-false value. We were using the false value in some situations to generate the sdp with `a=inactive`. This was specifically done on cases where an invite without sdp was sent and the user wanted audio, but got an audio video answer back. The result is now that the sdp generated has `a=recvonly` for video. Technically correct, but causes funky behavior for some endpoints. I will write a modifier to fix all of this.

I have some other bugs that have come up pending. I will try and get all of that pushed here tomorrow.

egreenmachine:
It has been brought to my attention that we need to handle retransmissions for renegotiation to work correctly. This is that SIP.js should eat retransmissions for messages that it already received and are the same. 

As a first pass to handle this I am going to write a simple hash table after the message is parsed based on some set of unique parameters to that message (call-id, from-tag, to-tag)? I will look at what some other SIP signaling libraries do about this (OpenSIPs and FreeSWITCH come to mind). Maybe a second PR to handle the retransmissions is a good start and then merge this after as this PR is getting a bit unwieldy.

egreenmachine:
After some discussion yeseterday, we realized that if we track messages in the dialog by `cseq` minus some edge cases (such as ACK) we can track retransmissions via `cseq`. Additionally on receiving a retransmission for something we have already replied to we need to retransmit the response. This should not be all that difficult to do. 

In terms on renegotiating, there should be some sort of indication to the client that a renegotiation has occurred. This all relates to the session description handler events that I have ignored for the time being.

egreenmachine:
So I went ahead and broke out the WebRTC modifiers into their own file. They can be used by passing it to the `SessionDescriptionHandlerFactory` for use on all calls, or on `invite` and `accept` calls for individual calls. This will help standardize the way we handle this type of this and allow them to be used elsewhere. I will document this a bit further on how to do it with the default WebRTC Session Description Handler.

egreenmachine:
# Download and Build
```
git clone https://github.com/onsip/SIP.js.git
cd SIP.js
git checkout session-description-rewrite
npm install
npm run buildAndTest
```

src/WebRTC/SessionDescriptionHandler.js
```diff
@@ -0,0 +1,487 @@
...
+  createOfferOrAnswer: {writable: true, value: function createOfferOrAnswer (RTCOfferOptions, modifiers) {
+    var self = this;
+    var methodName;
+    var pc = this.peerConnection;
+
+    RTCOfferOptions = RTCOfferOptions || {};
+
+    methodName = self.hasOffer('remote') ? 'createAnswer' : 'createOffer';
+
+    return pc[methodName](RTCOfferOptions)
```
egreenmachine: We should check the state of the peer connection signaling state to make sure that the peer connection is in the correct state before doing createOffer/createAnswer. Firefox will throw an error if you are not in the correct state, chrome allows the call to still succeed.

src/WebRTC/SessionDescriptionHandler.js
```diff
@@ -0,0 +1,487 @@
...
+   * @param {Object} [options.peerConnectionOptions] If this is set it will recreate the peer connection with the new options
+   * @param {Array} [modifiers] Array with one time use description modifiers
+   * @returns {Promise} Promise that resolves once the description is set
+   */
+  setDescription: {writable:true, value: function setDescription (sessionDescription, options, modifiers) {
+    var self = this;
+
+    // https://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browser
+    var isFirefox = typeof InstallTrigger !== 'undefined';
+    if (!this.session.disableRenegotiation && isFirefox && this.peerConnection && this.isVideoHold(sessionDescription)) {
```
egreenmachine: Discussion with FF team regarding this: https://groups.google.com/forum/#!topic/mozilla.dev.media/BDnRhxZsx6I

src/WebRTC/SessionDescriptionHandler.js
```diff
@@ -0,0 +1,487 @@
...
+          body: sdp,
+          contentType: self.CONTENT_TYPE
+        };
+      });
+    }
+
+    // GUM and set myself up
+    self.logger.log('acquiring local media');
+    // TODO: Constraints should be named MediaStreamConstraints
+    return this.acquire(self.constraints)
```
vadirn: what about providing already acquired mediastream?

egreenmachine: You will need to write your own sessionDescriptionHandler to handle that. It is not something we want to deal with on the default implementation.

egreenmachine:
# Simple

So I have added a `Simple` layer. For now, I am providing documentation here. Eventually it will need to be added to the readme and website.

### What is Simple?
`Simple` is a layer of abstraction on top SIP.js that aims to get new users up and running with SIP.js as quickly and easily as possible. It can make and receive calls, while updating a UI. It is only for use in a browser with WebRTC. 

## API

### Constructor
`var simple = new SIP.WebRTC.Simple(options);

#### options

```

**Comment by falconsapsan** [comment]
i came here with problem https://github.com/sipwise/rtpengine/issues/393
I looked this branch, but i got error.. 

Google Chrome | 59.0.3071.86  (64 bit)
sip.invitecontext.sessionDescriptionHandler | unable to acquire streams
NavigatorUserMediaError {name: "DevicesNotFoundError", message: "", constraintName: ""}

In Firefox 52.0.2 (32-bit)
TypeError: window.navigator.getUserMedia is not a function

If i use master branch - i'm don't got this errors

**Comment by egreenmachine** [comment]
@falconsapsan,

Your browsers are a bit out of date. Chrome should be 60+, and FF is 55+. This branch definitely will not work on your version of Firefox.

That being said, it appears that Chrome cannot access your media devices. Could you provide a [gist](https://gist.github.com/) of more detailed logging after updating your browsers?

**Comment by falconsapsan** [comment]
Chorme  61.0.3163.79, (64 bit)
https://gist.github.com/vectorcode/b932ac9ad76c9e7630e849093abf1e07

**Comment by egreenmachine** [comment]
@falconsapsan 

In your UA configuration can you try adding the following

**Comment by egreenmachine** [comment]
# Simple

So I have added a `Simple` layer. For now, I am providing documentation here. Eventually it will need to be added to the readme and website.

### What is Simple?
`Simple` is a layer of abstraction on top SIP.js that aims to get new users up and running with SIP.js as quickly and easily as possible. It can make and receive calls, while updating a UI. It is only for use in a browser with WebRTC. Currently it only allows for a single call to happen at any given time.

## API

### Constructor
`var simple = new SIP.WebRTC.Simple(options);`

#### options
At least one remote media element is required.

```
{
  media: {
    remote: {
      audio: <DOM element>,
      video: <DOM element>
    },
    local: {
      video: <DOM element>
    }
  },
  ua: {
    wsServers: <WebSocket Server>,
    uri: <SIP URI>,
    authorizationUser: <SIP Auth User>,
    password: <SIP Password>
  }
}
```

### Methods

#### call

Start a new call to the uri provided.
```
simple.call('hello@onsip.com');
```

#### answer

Answer an incoming call that is ringing.
```
simple.answer();
```

#### hangup

Hangup a call.
```
simple.hangup();
```

#### mute

Mute the local media of a call.
```
simple.mute();
```

#### unmute

Unmute the local mdia of a call.
```
simple.unmute();
```

#### hold

Place a call on hold and mute the local media.
```
simple.hold();
```

#### unhold

Take a call off hold and unmute the local media.
```
simple.unhold();
```

#### sendDTMF

Send a DTMF event via SIP Message
```
simple.sendDTMF('1');
```

## Events

Events are emitted the same in both incoming and outgoing calls.

#### registered

Emitted when the UA registers.

#### unregistered

Emitted when the UA unregisters.

#### new

Emitted when there is a new call. Before a progress message is sent or received.

#### connecting

Emitted once a progress message is sent or received.

#### connected

Emitted when the call is connected.

#### ended

Emitted when the call is completed, either from a hangup or an error.

#### hold

Emitted when a call is placed on hold or you place a call on hold.

#### unhold

Emitted when a call is taken off hold.

#### mute

Emitted when the mute function is called.

#### unmute

Emitted when the unmute function is called.

#### dtmf

Emitted when a SIP DTMF message is received

## States

Simple has 5 states that it can be in at any given time. The states will always flow in a single direction from `STATUS_NEW` to `STATUS_COMPLETED`. For example you cannot go from `STATUS_CONNECTED` to `STATUS_CONNECTING`, but you could go directly from `STATUS_CONNECTING` to `STATEUS_COMPLETED`. Once you are in `STATUS_COMPLETED` you can transition back to `STATUS_NEW`. 

#### STATUS_NULL

A blank state for when Simple is instantiated.

#### STATUS_NEW

A new call has been initialized.

#### STATUS_CONNECTING

The call is setting up, but is not yet finalized.

#### STATUS_CONNECTED

The call is now connected and media is flowing.

#### STATUS_COMPLETED

The call is completed.

**Comment by falconsapsan** [comment]
```
        let options =  {
            media: {
                remote: {
                    audio: document.getElementById('audio'),
                    video: ''
                },
                local: {
                    video: ''
                }
            },
            ua: {
                traceSip: false,
                displayName: 'Meee',
                uri: '103@1.sips.me',
                wsServers: ['wss://1.sips.me:7443'],
                authorizationUser: '',
                password: 'хххххххх',
                sessionDescriptionHandlerFactoryOptions: {
                    constraints: {
                        audio: true,
                        video: false
                    },
                    rtcConfiguration: {
                        rtcpMuxPolicy: 'negotiate'
                    },
                }
            }
        }
        this.userAgent = new SIP.WebRTC.Simple(options);
```

make call..
` this.userAgent.call('1003@1.sips.me');`

try Hold
`this.userAgent.hold();
`
try unHold
`this.userAgent.unhold();
`
103 don't hear 1003
103 - Google Chrome | 61.0.3163.91
1003 - PhonerLite user

Browser console - https://gist.github.com/falconsapsan/b230856a34c500da6d3111c25a39add7

**Comment by egreenmachine** [comment]
@falconsapsan I cannot reproduce. Can you check that your media server is behaving as expected?

**Comment by egreenmachine** [comment]
## Some notes on Safari (as of Tech Preview Release 39).

The media devices are chosen as the system defaults. The user has no option when allowing media to choose a different set of devices, like in Chrome or Firefox.

Safari does not allow the auto-playing of a media element from an async javascript function. The play function must be called from a synchronous user action. This is a bit annoying, but to get around it you can set the autoplay attribute on the media element that you are trying to do. This is pretty well documented as a Safari 'quirk' and documentation around the internet on this is pretty good.

On my Late 2016 MacBook Pro, when Safari requests the microphone it puts the internal microphone in 4-channel 24-bit audio mode. Why does it need this many channels? I could not say. However Firefox cannot handle the internal Macbook Pro microphone being in this mode. The error message that it spits out is extremely vague at best. 

```
MediaStreamError { name: "NotReadableError", message: "Failed to allocate audiosource", constraint: "", stack: "" }
```

The way to fix this is to go into `Applications > Utilities > Audio Midi Setup` and choose the Built-in Microphone from the left panel. Then under format change it to 2 ch. Firefox will then work again, but if you use Safari again it will switch it back on it's own. I imagine that this will frustrate a lot of Macbook owners, but I have not tested other Mac models.

**Comment by egreenmachine** [comment]
Assuming that there are no glaring issues, I am starting on documentation with the expectation to merge this and do a version bump first thing tomorrow.

**Comment by schirrmie** [comment]
Hi,
i didn't know how to get the console output in mobile, but now I have the log and see this:

```
Mon Nov 20 2017 20:04:46 GMT+0100 (CET) | sip.invitecontext.sessionDescriptionHandler | acquiring local media
Mon Nov 20 2017 20:04:46 GMT+0100 (CET) | sip.invitecontext.sessionDescriptionHandler | unable to acquire streams
NotSupportedError: Only secure origins are allowed (see: https://goo.gl/Y0ZkNV).

```
So I have to use https, now its works! Thank You!
Sorry for my stupid problems.

PR #430, opened 2017-07-06
Set custom username in contact's user part

### LINKED ISSUE #428: Set custom username in Contact's user part
Would it be possible to add option to send `settings.authorizationUser` instead of `SIP.Utils.createRandomToken()` in `Contact` ?

instead of
```js
...
uri: new SIP.URI('sip', SIP.Utils.createRandomToken(8), settings.viaHost, null, {transport: settings.contactTransport}),
...
```
to have this

```js
...
uri: new SIP.URI('sip', settings.authorizationUser, settings.viaHost, null, {transport: settings.contactTransport}),
...
```

With settings option:

```js
const user = settings.addRandomUser ? SIP.Utils.createRandomToken(8) : settings.authorizationUser;
...
uri: new SIP.URI('sip', user, settings.viaHost, null, {transport: settings.contactTransport}),
...
```

VadimDez:
Closes #428

### DIFF
--- a/src/UA.js
+++ b/src/UA.js
@@ -963,7 +963,10 @@ UA.prototype.loadConfig = function(configuration) {
         return new SIP.DigestAuthentication(ua);
       }),
 
-      allowLegacyNotifications: false
+      allowLegacyNotifications: false,
+
+      // user name in user part
+      userName: null
     };
 
   // Pre-Configuration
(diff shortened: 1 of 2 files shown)


james-criscuolo:
Thanks for the PR! Given that we have a `contactTransport` config variable, I think `contactName` would be a bit more informative here than `userName`. I would also move this up next to `contactTransport`, and set the default to the randomToken instead of null. At that point, you could get rid of the `user` variable and just pass `settings.contactName`.

Lastly, there should also be a test to check the default case still works (although that may already exist).

james-criscuolo:
This looks good now, thanks! I'll give it a few days for other authors and contributors to take a look before merging.

james-criscuolo:
I'm merging now, thanks again

PR #466, opened 2017-11-08
Check whether constraints have changed before calling getUserMedia

jlaine:
The "if" condition could never be true, as we don't ever receive an
"options.constraints" object which *is* (reference equality) this.constraints.

Instead, prepare the target constraints by merging the current and new
constraints, and do a deep comparison with what we currently have.

### DIFF
--- a/src/WebRTC/SessionDescriptionHandler.js
+++ b/src/WebRTC/SessionDescriptionHandler.js
@@ -113,13 +113,14 @@ SessionDescriptionHandler.prototype = Object.create(SIP.SessionDescriptionHandle
       this.initPeerConnection(options.peerConnectionOptions);
     }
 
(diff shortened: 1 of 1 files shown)


egreenmachine:
We like what you did with `Object.assign` and assigning the passed constraints over the default constraints then comparing. We should take this approach over all the calls into the Session Description Handler, and then we can call `checkAndDefaultConstraints` in just one place, right before we call `getUserMedia`. This is not necessary for this PR, but we figured it was worth noting for future cleanup.

Overall this looks good on our end. We are going to let it sit for a few days to get feedback from the community. The plan is to merge it before our next release (barring any issues cropping up). 

Thanks for the contribution.

jlaine:
Awesome, thanks Eric!

PR #467, opened 2017-11-08
Improve detection of ICE gathering complete

jlaine:
The previous detection mechanism had several issues:

- it also checked the iceConnectionState, which is unrelated

- it did not account for the fact iceGatheringState can go from
  'complete' back to 'gathering'

- it did not ensure the timer was stopped when destroying the
  session description handler

This also adds unit tests for the ICE gathering state changes.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -13,7 +13,7 @@ module.exports = function(config) {
       'test/polyfills/*.js',
       'test/helpers/*.js',
       'dist/sip.js',
-      'test/spec/*.js'
+      'test/spec/**/*.js'
     ],
 
     // list of files to exclude
@@ -21,7 +21,9 @@ module.exports = function(config) {
 
     // preprocess matching files before serving them to the browser
     // available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
-    preprocessors: {},
+    preprocessors: {
+      'test/spec/WebRTC/SessionDescriptionHandler.spec.js': 'webpack'
+    },
 
     // test results reporter to use
     // possible values: 'dots', 'progress'
@@ -62,10 +64,11 @@ module.exports = function(config) {
     concurrency: Infinity,
 
     plugins : [
-	    'karma-jasmine',
+      'karma-jasmine',
       'karma-jasmine-html-reporter',
       'karma-mocha-reporter',
-	    'karma-phantomjs-launcher'
-	  ]
+      'karma-phantomjs-launcher',
+      'karma-webpack'
+    ]
   })
 }
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "karma-jasmine-html-reporter": "^0.2.2",
     "karma-mocha-reporter": "^2.2.5",
     "karma-phantomjs-launcher": "^1.0.4",
+    "karma-webpack": "^2.0.6",
     "pegjs": "^0.10.0",
     "pegjs-loader": "^0.5.4",
     "uglifyjs-webpack-plugin": "^1.0.1",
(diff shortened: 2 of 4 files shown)


egreenmachine:
I would prefer if we emit the session description handler instead of the peer connection (revert d47d62e). We should try and be consistent and any event emissions from the session description handler should pass the session description handler itself as an argument. (Cleaning up events is on my todo list).

The other thing I would like to see is the `defer` removed. It is an old way of doing promises, and I do not mind keeping around code that already uses it but since this code is being touched anyways we should put the effort in to remove it. As far as I can tell, there is no reason that there should be an array of deffereds. It can be a single promise that is cleared and reset once ice completes, so that it is ready to go next time around. 

Barring any other issues that crop up, once those 2 things are addressed I think that this is good to merge. This PR looks well thought out and we appreciate the community support.

jlaine:
Emitting the session handler: ok

Using a single promise doesn't seem right, this is a big part of what is being addressed: we don't have any guarantees that there is only a single invocation of waitForIceGarheringComplete waiting for ice gathering to complete. We are using a single event handler to resolve/reject 0-N promises, the array of deferred does exactly that. If you *want* it only to be possible to have one, we need to reject the previous promise when a new one is created.

jlaine:
As for not using defer (i.e. keep resolve/reject to a promise) the way to do that is to stop using onicegaheringstatechange and use addEventListener instead. I am not sure this is what we want though, as to be fully correct we should reject any outstanding promises when the peer connection changes.

egreenmachine:
I wrote this [fiddle](https://jsfiddle.net/d9wfajjL/1/) to show how you can have more than 1 `then` on a single promise. I believe that is what you are concerned about. 

In terms of the deferred there are a few options to get rid of it in the current form. They are a bit difficult to explain, so I can just clean it up after pulling. The array of defers is the bigger issue.

But yes, the promise should probably reject if it is still outstanding and the peer connection changes.

jlaine:
My point is not so much that you can have multiple then() on a promise : that much is clear. My point is that in getDescription() what we want is : "notify me the next time ICE gathering completes (or times out)". So consecutive requests to what is now called "waitForIceGatheringComplete" are different promises : they depend on the point in time at which they were formulated.

An alternative implementation is to go back to the single deferred, but explicitly reject + reinitialise it whenever  :

a/ the peerConnection changes
or
b/ the ice gathering state goes back to "gathering"

egreenmachine:
Yes, I think what you stated as "an alternative" is exactly what I was thinking.

As the code stands in this PR... If you were to `waitForIceGatheringComplete` it completes, then changes back to gathering, the changes back to complete. The promise that was returned when you called `waitForIceGatheringComplete` would not alert of state change again. You are building an array of promises, but treating them as a single promise, resolving and clearing the array all at once regardless. I hope that makes sense.

jlaine:
The whole point of a promise is that it resolves only once, which is precisely what we want in getDescription: for a given call to getDescription the code in this PR will resolve whenever we next reach the "complete" state (or are already in complete, or time out). At this point any outstanding promises are flushed from iceGatheringDeferred.

If we go back to "gathering": no problem, future calls to getDescription will also correctly wait for the gathering state to stabilise. If you want I would be happy to provide a test which shows the code is correct.

I think maybe a short chat / call would be a good idea, as the idea here is to get things absolutely right. Feel free to email me to arrange this.

jlaine:
The pull request has now reached a state I am entirely satisfied with, the unit tests were really necessary as they shook out several cases I had not initially thought of. Let me know if you have any additional concerns.

PR #468, opened 2017-11-08
Bind holdModifier to the SessionDescriptionHandler instance

jlaine:
When putting a call on hold, sessionDescriptionHandler.holdModifier is
added to the modifiers. The modifier is an instance method, and as such
we need to bind it to the instance.

It so happens the default WebRTC holdModifier does not reference 'this',
so things work as expected. However for a custom session description
handler this may not hold true.

### DIFF
--- a/src/Session.js
+++ b/src/Session.js
@@ -416,7 +416,7 @@ Session.prototype = {
 
     options = options || {};
     options.modifiers = modifiers || [];
-    options.modifiers.push(this.sessionDescriptionHandler.holdModifier);
+    options.modifiers.push(this.sessionDescriptionHandler.holdModifier.bind(this.sessionDescriptionHandler));
 
     this.local_hold = true;
 


egreenmachine:
So, I think that modifiers should be static instead of instance methods. This is what we did for other modifiers, this is the only one we put as an instance method.

The only argument we could see for having it as an instance method is so that the modifier can change it's own behavior based on the state of the session. We think that this should be left up to whatever is adding the modifiers and different modifiers should be added based on the state and the modifiers should be static. 

So the correct thing here is to change the hold modifier to be a static method. I can get this in soon, or you can create another PR that does this. Either way I am going to close this PR, but feel free to continue the conversation here.

jlaine:
OK, I don't have any strong feelings about this, other than the fact it's not documented. I just tripped over this by trying to add a this.logger.log(..) call inside holdModifier

egreenmachine:
I spent a bit of time looking at this further. I think that putting the hold modifier on the class makes the entire code more complex, especially with defining custom session description handlers. So the hold modifier here will stay as an instance method, but behave as a class method. That is how it is written now, and I will update the documentation to say as much.

PR #469, opened 2017-11-13
Fix G.722 stripping of non-audio call

rscreene:
When making a video only call using the Simple API I was getting a error when stripping the G.722 codec ('cos it was on a Mac).
This PR checks that we've got a audio mline before referencing it.


egreenmachine:
I caught this at the end of the day Friday and was going to try and address it today. Thanks for this.

PR #475, opened 2017-11-15
Consistently use iceCheckingTimeout = 5000 as default value

jlaine:
Currently, UA uses a sensible default iceCheckingTimeout: 5000.

However, as Simple overrides sessionDescriptionHandlerFactoryOptions, it loses
this default value and as a result does not use any iceCheckingTimeout.

This also adds a test to check the default options used by Simple.

### DIFF
--- a/src/UA.js
+++ b/src/UA.js
@@ -932,10 +932,7 @@ UA.prototype.loadConfig = function(configuration) {
       // Session Description Handler Options
       sessionDescriptionHandlerFactoryOptions: {
         constraints: {},
-        peerConnectionOptions: {
-          iceCheckingTimeout: 5000,
-          rtcConfiguration: {}
-        }
+        peerConnectionOptions: {}
       },
 
       contactName: SIP.Utils.createRandomToken(8), // user name in user part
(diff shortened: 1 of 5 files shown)


jlaine:
NOTE : the same issue affects any UA which is created with a sessionDescriptionHandlerFactoryOptions. An alternative fix would be to move the default iceCheckingTimeout *inside* WebRTC/SessionDescriptionHandler, and ask users to explicitly set iceCheckingTimeout to 0 if they want to disable it.

What's your preference Eric?

egreenmachine:
> An alternative fix would be to move the default iceCheckingTimeout inside WebRTC/SessionDescriptionHandler, and ask users to explicitly set iceCheckingTimeout to 0 if they want to disable it.

That seems ideal here

jlaine:
In that case I think we can collapse peerConnectionOptions to {}:

https://github.com/onsip/SIP.js/blob/49c6ac45b673e9f8c490b8aa27041eb70ace8d48/src/UA.js#L935

rtcConfiguration is initialised to {} if absent:

https://github.com/onsip/SIP.js/blob/49c6ac45b673e9f8c490b8aa27041eb70ace8d48/src/WebRTC/SessionDescriptionHandler.js#L347

I haven't checked but we can possibly even remove peerConnectionOptions altogether, and possibly constraints too.

jlaine:
Woops, not good, I'm adding the option in the wrong place

jlaine:
OK, fixed, it now works as expected

PR #488, opened 2017-11-27
Use babel-loader to transpile code

jlaine:
Switch from eslint-loader to babel-loader to allow using modern constructs
such as arrow functions or ES6 classes while still retaining "compatibility"
with older browsers. It's worth mentioning browsers which support WebRTC
already ship with ES6 support, but at least we won't trigger parser errors
on the older browsers.

The cost in terms of runtime size is negligeable, the minified code grows
by less than 400 bytes as a result of the switch. Note: we don't pull in
any polyfills with this commit, so only ES6 constructs are supported, not
using new classes such as Map or Set.

We still perform linting as part of the build, by invoking eslint in the
prebuild step.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,8 +26,10 @@
     "javascript"
   ],
   "devDependencies": {
+    "babel-core": "^6.26.0",
+    "babel-loader": "^7.1.2",
+    "babel-preset-env": "^1.6.1",
     "eslint": "^4.9.0",
-    "eslint-loader": "^1.9.0",
     "jasmine-core": "^2.8.0",
     "karma": "^1.7.1",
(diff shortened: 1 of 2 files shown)


james-criscuolo:
This looks good to me. Eslint still runs automatically while building, and it failing still halts the build with an error.

Generally speaking, longterm we would like to move to typescript, but babel is an adequate and noninvasive middle-step.

jlaine:
Any chance of merging this @egreenmachine ?

egreenmachine:
My goal is to do a `0.9.0` release with #459 which was just merged, then we shall see what is next. It is in the queue.

PR #502, opened 2018-01-09
Support for early media (with 100rel or not) for InviteClientContext

mocodo:
Hi,

We tried to make early media work and we managed to get something that seems to work. It might not be the most elegant solution.

### DIFF
--- a/src/Session.js
+++ b/src/Session.js
@@ -1522,7 +1522,7 @@ InviteClientContext.prototype = {
           }
 
           // TODO: This may be broken. It may have to be on the early dialog
-          if (!this.ua.configuration.allowEarlyMedia) {
+          if (!this.sessionDescriptionHandler) {
             this.sessionDescriptionHandler = this.sessionDescriptionHandlerFactory(this, this.sessionDescriptionHandlerFactoryOptions);
           }
 
--- a/src/Session.js
+++ b/src/Session.js
@@ -1521,7 +1521,10 @@ InviteClientContext.prototype = {
             return;
           }
           // TODO: This may be broken. It may have to be on the early dialog
-          this.sessionDescriptionHandler = this.sessionDescriptionHandlerFactory(this, this.sessionDescriptionHandlerFactoryOptions);
+          // The following if statement may solve the issue
+          if (!this.sessionDescriptionHandler) {
+            this.sessionDescriptionHandler = this.sessionDescriptionHandlerFactory(this, this.sessionDescriptionHandlerFactoryOptions);
+          }
           if (!this.sessionDescriptionHandler.hasDescription(response.getHeader('Content-Type'))) {
             extraHeaders.push('RAck: ' + response.getHeader('rseq') + ' ' + response.getHeader('cseq'));
             this.earlyDialogs[id].pracked.push(response.getHeader('rseq'));
@@ -1589,7 +1592,21 @@ InviteClientContext.prototype = {
             });
           }
         } else {
-          this.emit('progress', response);
+          // Early media
+          if (this.hasOffer && response.status_code === 183) {
+            this.hasAnswer = false;
+            this.sessionDescriptionHandler.setDescription(response.body, this.sessionDescriptionHandlerOptions, this.modifiers)
+              .then(function onSuccess() {
+                session.status = C.STATUS_EARLY_MEDIA;
+                session.emit('progress', response);
+              }, function onFailure(e) {
+                session.logger.warn(e);
+                session.acceptAndTerminate(response, 488, 'Not Acceptable Here');
+                session.failed(response, SIP.C.causes.BAD_MEDIA_DESCRIPTION);
+              });
+          } else {
+            this.emit('progress', response);
+          }
         }
         break;
       case /^2[0-9]{2}$/.test(response.status_code):
--- a/src/Session.js
+++ b/src/Session.js
@@ -1520,11 +1520,12 @@ InviteClientContext.prototype = {
               (this.earlyDialogs[id].pracked[this.earlyDialogs[id].pracked.length-1] >= response.getHeader('rseq') && this.earlyDialogs[id].pracked.length > 0)) {
             return;
           }
+
           // TODO: This may be broken. It may have to be on the early dialog
-          // The following if statement may solve the issue
-          if (!this.sessionDescriptionHandler) {
+          if (!this.ua.configuration.allowEarlyMedia) {
             this.sessionDescriptionHandler = this.sessionDescriptionHandlerFactory(this, this.sessionDescriptionHandlerFactoryOptions);
           }
+
           if (!this.sessionDescriptionHandler.hasDescription(response.getHeader('Content-Type'))) {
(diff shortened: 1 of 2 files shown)


egreenmachine:
The problem with this code is that it will not function correctly if the call forks. I do not think I can merge this in good faith knowing that it will break in RFC compliant call scenarios.

mocodo:
So currently there is no way we can get early media support ? 

In addition, when 100rel is set to SUPPORTED or REQUIRED on user agent configuration, we have no media at all (inbound/outbound). 

Is there a way we could help with this ?

egreenmachine:
You need to do 100 rel and invite without sdp. I believe that early media works in this scenario. If it doesn't we can look into that.

mocodo:
Unfortunatly, our servers won't let us invite without SDP. I think we'll have to go deeper on this patch to make it work :)

onsip:
Charly,

What server are you working with if you don't mind?

Best,

[image: Mike Oeth] Mike Oeth
CEO & Co-Founder
O:  215 701 3051 <%28215%29%20701-3051> | C:  215 630 3498
<%28215%29%20630-3498> |  1-800-801-3381, ext 7001
<https://insta.onsip.com/phone/call/?n=Mike%20Oeth&a=mike%40junctionnetworks.com>
[image: onsip] <http://www.onsip.com/> [image: facebook]
<https://www.facebook.com/onsip> [image: twitter]
<https://twitter.com/onsip> [image: linkedin_socialicon.png]
<https://www.linkedin.com/in/mikeoeth> [image: blog]
<http://www.onsip.com/blog>
------------------------------
OnSIP Direct Call Links:Sales
<https://app.onsip.com/app/call?a=acdsalesqueuebhr%40junctionnetworks.com&n=OnSIP+Sales>
,Support
<https://app.onsip.com/app/call?a=acdsupportqueuebhr%40junctionnetworks.com&n=OnSIP+Support>,
Billing
<https://app.onsip.com/app/call?a=billing.bhr%40junctionnetworks.com&n=OnSIP+Billing>

Have you tried the new OnSIP Desktop app
<https://support.onsip.com/hc/en-us/articles/217065557-The-OnSIP-App>?

egreenmachine:
The real issue is that this is a limitation of the WebRTC Peer Connection. We are applying the early media as a final response. This is because WebRTC does not support pre-answers. So in the case of a forked call...

Alice calls (with SDP) Bob. Bob has 2 endpoints Bob1 and Bob2. Both endpoints send early media. So now it is a race for whoever can get early media to Alice faster. Lets say that Bob1 gets early media to Alice first, the SDP from Bob1 is applied as the final answer to Alice's offer. Then Bob2 actually answers the call. There is no way for us to override the SDP that was already applied from Bob1. The call is dead.

Invite without SDP fixes this by reversing the SDP offer answer. So in the above example Bob1 and Bob2 are sending SDP offer's in the 183. This allows us within SIP.js to build a Peer Connection for each fork of the call, and generate distinct answers for each fork. When we get the final 200 Ok from any fork, we can simply discard the other peer connections.

mocodo:
It makes sense indeed, we do not have this kind of scenario, hence I didn't checked it. Thank you for this explanation.

egreenmachine:
Thinking about it further... maybe we could put this code behind a UA configuration option and document the broken forking scenario if the option is enabled. That might be a good middle ground to get your scenario working, while not breaking the rest of our user base.

mocodo:
I will try to submit a PR with this solution and we'll see if it fits :)

mocodo:
So I tried something, I can confirm it works on our scenario. Waiting for you to see if UA configuration naming suits the needs.

src/Session.js
```diff
@@ -1520,8 +1520,12 @@ InviteClientContext.prototype = {
               (this.earlyDialogs[id].pracked[this.earlyDialogs[id].pracked.length-1] >= response.getHeader('rseq') && this.earlyDialogs[id].pracked.length > 0)) {
             return;
           }
+
           // TODO: This may be broken. It may have to be on the early dialog
-          this.sessionDescriptionHandler = this.sessionDescriptionHandlerFactory(this, this.sessionDescriptionHandlerFactoryOptions);
+          if (!this.ua.configuration.allowEarlyMedia) {
```
egreenmachine: I do not think that this block of code is necessary. I think it is best to revert it to how it was previously.

mocodo: Must say I am confused, there is only one way `this.sessionDescriptionHandler` is going to be `null` at this point which is an INVITE without SDP. Creating a new sessionDescriptionHandler is going to initiate a new peerConnection, different from the one used in my INVITE request no?

egreenmachine: What I am saying is that the block is not necessary and potentially sets up an undefined variable error on line 1529 below. It is possible to get to this code and have the allowEarlyMedia flag be `true`. That would cause the SDH to not get instantiated and then an undefined error to spit out on the next `if` statement.

mocodo: I see, then an `if` statement checking if `this.sessionDescriptionHandler` is defined should do the trick right ?

egreenmachine: Yes. That would work fine.

egreenmachine:
So having some further discussions, we think that this should actually send cancels to any branches that come in after the first 183 comes in the case of invite without sdp and the option being enabled. It looks like sending cancels to individual branches is RFC compliant, we are just not sure if SIP.js is currently set up to do this. We are going to look into it further, but I wanted to keep you in the loop. I am about to do a release and this will not make it in because of that. Thanks!

GMLudo:
Hi,

We did some tests of this pull request on 0.11, it seems to resolve our problem of early media.
Do you plan to merge this pull request or propose an alternative way for that ?

Regards.

egreenmachine:
If the PR is updated to send cancels to all other pending branches, we would take this. This is not something that the maintainers are actively working on at the moment.

tm1000:
@egreenmachine without looking through all the code base would you be able to give a hint of where to start on such a thing?

tm1000:
FYI this fails in random cases on Chrome but not on Firefox with

```
DOMException: Failed to set remote answer sdp: Called in wrong state: kStable
```

egreenmachine:
As I have stated previously there are definitely some issues with this PR. It is not high on my priority list, but if you can get it to a good place we would take it.

Zwoniar:
Is there any changce to merge this? Cause it's working for me too :D

egreenmachine:
@Zwoniar If you can show this branch working with call forking we will merge it.

TayyabTalha:
On Centos 7 and Asterisk 15.4.1 after making these changes Unhold is not working any solution for that?

Jaykah:
@mocodo did you ever get that working with call forking?

Jaykah:
@egreenmachine if we do the rework, would you merge it? or is there some work going on at that front already?

thanks!

james-criscuolo:
Hi @Jaykah,
  This case is actually going to be covered in our next version with our new API, which can be seen in PR #705. We are looking to get a bit more documentation before merging, but we are expecting to release within the next week or two.

PR #507, opened 2018-01-18
Prefer ontrack over onaddstream, remove onremovestream

jlaine:
Until now, we used *both* the spec-compliant ontrack and the deprecated
onaddstream APIs to detect track/stream addition. This resulted in a
warning even on spec-compliant browsers which implement ontrack.

We now detect support for ontrack and only use onaddstream as a fallback,
and emit a more explicit warning message, formatted like the two other
warnings about deprecated WebRTC APIs.

We also remove the deprecated onremovestream which did nothing except
print a logging message.

### DIFF
--- a/src/WebRTC/SessionDescriptionHandler.js
+++ b/src/WebRTC/SessionDescriptionHandler.js
@@ -370,21 +370,18 @@ SessionDescriptionHandler.prototype = Object.create(SIP.SessionDescriptionHandle
     this.logger.log('New peer connection created');
     this.session.emit('peerConnection-created', this.peerConnection);
 
-    this.peerConnection.ontrack = function(e) {
-      self.logger.log('track added');
-      self.emit('addTrack', e);
-    };
-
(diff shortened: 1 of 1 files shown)


james-criscuolo:
Thanks for the PR! This looks good. I'm going to leave this for a few days in case others have thoughts, and then I'm tempted to leave this as a pull request until we are ready to release 0.10.0, as its an outward-facing API change. It's possible that our next version is 0.10.0, but I think leaving this gives us more flexibility.

Thanks,
James

jlaine:
Sure, there's no great rush.

I've tested this change on Firefox (spec compliant), Chrome (lacks ontrack) and Chrome + webrtc-adapter (behaves as if spec-compliant).

jlaine:
As a side note : Chrome 64 includes the getSenders() and ontrack APIs. This means all major browsers will finally be using spec-compliant APIs!

https://bugs.chromium.org/p/chromium/issues/detail?id=760107

jlaine:
Ping?

PR #509, opened 2018-01-30
add support for RTCDTMFSender

JimGreenberg:
add support for https://developer.mozilla.org/en-US/docs/Web/API/RTCDTMFSender which will be used for DTMF in favor of the SIP packet approach as is currently implemented. If possible, we access the exposed RTCRtpSender.dtmf property- and if that fails we fall back to calling RTCPeerConnection.createDTMFSender() passing in an audio track and if that also fails we fall back to the previous implementation.

tested on chrome 63/64 as well as firefox

### DIFF
--- a/src/Session.js
+++ b/src/Session.js
@@ -30,6 +30,7 @@ Session = function (sessionDescriptionHandlerFactory) {
   this.dialog = null;
   this.pendingReinvite = false;
   this.earlyDialogs = {};
+  this.dtmfSender = null;
   if (!sessionDescriptionHandlerFactory) {
     throw new SIP.Exceptions.SessionDescriptionHandlerMissing('A session description handler is required for the session to function');
   }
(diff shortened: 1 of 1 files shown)


james-criscuolo:
The only thing I'd consider altering here is the logging. In the old system, with traceSip on, we'd get a large set of logs informing us that we sent DTMF. In the new system, I do not believe there is any.

I think an info level log probably makes sense per dtmf input.

egreenmachine:
Spoke about this offline briefly this morning. To summarize my thoughts on this.

1. This behavior needs to be configurable for the user. 
2. The default should be doing what SIP.js currently does, with the expectation that we will move to this stuff by default in the future.
3. Logging about which way DTMF is being sent, warning the user about the default changing, and giving info about what DTMF is happening.
4. Anything related to the `peer connection` cannot be inside the `Session` object. This will have to move into the `Session Description Handler` and new methods on the SDH exposed.
5. Just a bit more robust guarding against potential crashes.
6. Unit testing is always appreciated!

PR #516, opened 2018-02-19
Make SIP.js (hopefully) buildable on Windows

seanbright:
If eslint requires LF, make sure JS files get checked out that way

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,4 @@
 *.min.js binary
+
+# eslint expects all JS files to have Unix line endings
+*.js text eol=lf
--- a/package.json
+++ b/package.json
@@ -29,6 +29,8 @@
     "babel-core": "^6.26.0",
     "babel-loader": "^7.1.2",
     "babel-preset-env": "^1.6.1",
+    "cpr": "^3.0.1",
+    "cross-env": "^5.1.3",
     "eslint": "^4.9.0",
     "jasmine-core": "^2.8.0",
     "karma": "^1.7.1",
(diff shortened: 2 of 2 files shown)


egreenmachine:
Hey @seanbright, thanks for the PR. I have outlined some of my thoughts below. We are developing in a UNIX environment so I am really not all that familiar with getting this to run on Windows.

I am uncomfortable with the `.gitattributes` change. I think it may cause adverse side effects particularly on the release branch with minified javascript. Based on my reading, it appears the correct solution is to set the attributes locally in your dev environment. 

In terms of the copy files, can you provide some evidence that this is a generally accepted approach? I have found a few different ways to achieve this and it appears that the most minimal and accepted way of doing this is to actually write a node script to copy the file and then execute that script with node, which we should expect to be available on all systems. The concern here is that we are using tools and an approach that is not really all that common and it could create more problems than it solves, especially in terms of future support.

seanbright:
> I am uncomfortable with the `.gitattributes` change. I think it may cause adverse side effects particularly on the release branch with minified javascript. Based on my reading, it appears the correct solution is to set the attributes locally in your dev environment.

I assume you mean by setting `core.autocrlf` but that is a global git configuration flag, not a per-repository one.

> In terms of the copy files, can you provide some evidence that this is a generally accepted approach?

I cannot. I just looked around for other people experiencing the same problems with Unix specific commands and environment variable and determined what their solution was.

seanbright:
> I am uncomfortable with the .gitattributes change. I think it may cause adverse side effects particularly on the release branch with minified javascript.

If this is something that I can assist with testing, I am more than willing.

egreenmachine:
Hi @seanbright,

While we appreciate the effort to make this work, we are not going to take this PR. You can clone the repo with `--config core.autocrlf=input` which will fix the crlf issue. Additionally, you can compile SIP.js using git bash on windows without adding further dev dependencies to our repo.

We looked at other (bigger) projects on Github to see what they were doing in regards to these things cross platform and they are not addressing it in this matter. As far as I can tell the recommended solutions are to clone with the config flag or change your core git config. As for the cp, the recommended solutions is to use git bash.

seanbright:
Hey @egreenmachine,

Was there anything done on your side to see if the `.gitattributes` change would actually have any negative impact on your release process or are you just assuming that it would?

> As for the cp, the recommended solutions is to use git bash.

Recommended by whom?

egreenmachine:
We did not do any testing to see if the change would have a negative impact. My assumption currently is that there would be little to no impact, but it is not something that we need to do. The issue is really outside the scope of this project.

It is a generally recommended approach for getting a UNIX like shell on PC. Similar to the CRLF issue, we believe that getting `cp` to work on a system is outside the scope of this project.

seanbright:
I tested the `--config core.autocrlf=input` flag on clone and that seemed to do the trick. The bug about using the `cp` command instead of a webpack plugin will just have to remain outstanding.

> The issue is really outside the scope of this project.

I am not sure that I consider build failures on non-Unixy platforms "out of scope" for a project that assembles a handful of smaller text files into one larger text file - especially considering that every component of your build toolchain supports cross platform development. We'll just have to agree to disagree.

PR #534, opened 2018-03-20
Emit "terminated" on renegotiation error (+ test fix)

nud:
Previously SIP.js would send a BYE but the client code would never be notified of that, so the UI would basically not be able to recover.

While I was at it I also fixed the WebRTC.Simple test on non-Safari browsers, to avoid keeping a false positive.


james-criscuolo:
Hi @nud,
Thank you for the PR! This looks good to me, I'll give it a few more days for comments (at minimum, so @egreenmachine can see it) before merging.

Thanks again,
James

PR #541, opened 2018-03-28
Remove redundant track during renegotiation

x-games:
After renegotiation by reinvite two audio tracks remains active, because it add new track to peerConnection but never removes the old one. This commit removes redundant track


egreenmachine:
This is a good first effort, but I believe that it will end up creating more problems than it fixes. What really needs to be done is that we need better control of when `acquire` is called. We should track whether the local media is active or the constraints for local media has changed, then call acquire accordingly. When we call acquire, all local tracks need to be removed from the peer connection then all the tracks returned from GUM added.

egreenmachine:
Closing in favor #571 which was released as part of 0.11.1.

PR #549, opened 2018-04-25
Add P-Asserted-Identity header support

dsandras:
RFC 3325 indicates:
Typically, a user agent renders the value of a P-Asserted-Identity
header field that it receives to its user.  It may consider the identity
provided by a Trust Domain to be privileged, or intrinsically more
trustworthy than the From header field of a request.  However, any
specific behavior is specific to implementations or services.

The header has been added so that UA's can use it instead of the
remoteIdentity if this behavior is required.

Polycom, SNOM, Bria, Ekiga                                                                                                                                            render the P-Asserted-Identity header if they receive it during a
reINVITE, or in a 1xx/2xx response.

### DIFF
--- a/src/ServerContext.js
+++ b/src/ServerContext.js
@@ -24,6 +24,9 @@ ServerContext = function (ua, request) {
 
   this.localIdentity = request.to;
   this.remoteIdentity = request.from;
+  if (request.hasHeader('P-Asserted-Identity')) {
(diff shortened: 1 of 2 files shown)


egreenmachine:
Looks good. I will leave it open for a few days to see if anyone else has comments then merge it assuming nothing comes up.

dsandras:
Thanks!

PR #551, opened 2018-05-04
Support for outbound PUBLISH requests (RFC3903)

andrew127:
This commit adds the support for SIP PUBLISH (SIP Extension for Event State Publication, as described in RFC3903). Currently, only outgoing publish requests are supported.

## SIP.PublishContext
Extends `SIP.ClientContext`.

### Construction
Publishers are obtained via the `SIP.UA` `publish` method.
```
var publisher = ua.publish(target, event, body[, options]);
```

### Instance Attributes

**event**
`string`, event to publish

**target**
`string`|`SIP.URI`, request target

**body**
 `string`, request body to send with publish
 
**options**
`object`, optional. Object with extra parameters, see below

**options.extraHeaders**
`array`, optional. Array of strings with extra SIP headers to add with publish/unpublish requests

**options.contentType**
`string`, optional. Content-Type header representing body content type. Default is 'text/plain'

**options.expires**
`int`, optional. Default expire value for the published event. If not set, default value 3600 is used

**options.unpublishOnClose**
`boolean`, optional. If set true, UA will gracefully unpublish for the event on UA close. Default value is true.

### Instance Methods

**publish([body])** 
publish the event state. Used for updating the current state for the event or to restore the publish after the unpublish was called. When no body provided, existing state is refreshed ( no need to do this manually, as state will be refreshed periodically on 0.9 * Expire intervals)

**unpublish()**
removes published event

**close()**
removes published event and UA internal pointer to the object

### Events

`published` Fired for a publish request success (2xx response received on PUBLISH request)
```
.on('published', function (response, cause) {});
```

`unpublished` Fired for a unpublish request success or for a periodic publish refresh failure
```
.on('unpublished', function (response, cause) {});
```

`progress` Fired for a publish/unpublish request progress. This event indicate session progress state (1xx response received) or recovering after the 412 or 423 responses
```
.on('progress', function (response, cause) {});
```

`failed` Fired for a publish request failure or on a periodic publish refresh failure
```
.on('failed', function (response, cause) {});
```


andrew127:
Hi Jim, thank you for the feedback.
Yes,  agree with all the points,  will rewrite the module.
As the PUBLISHes are out of dialog and do not create new dialog,  I'd propose to do a `PublishContext` extending `ClientContext` ( similar to `RegisterContext` ) exposing `publish`, `unpublish` and `close` methods.
As for new exception, can add the `MethodParameterError(method, parameter, value)` to `SIP.Exceptions` module.
And,  have noticed few other things that can be done better )

andrew127:
Have refactored the code:
- `removeOnClose` is now part of the `options`
- statuses were removed as unneeded
- added `MethodParameterError` to Exceprions.js
- `SIP.publish` returns now instance of `PublishContext`
- unit tests added
- some internal improvements and fixes
- description on the initial post updated to reflect the changes

JimGreenberg:
Hey Andrew, awesome work! I was curious as to why you implemented `PublishContext` the way you did, having it 'wrap' `ClientContext` instead of extending it, which seems simpler. Functionally, it's all there though.
The other small thing is that you have a typo in the filename `test/spec/SpecPublihsContext.js`
Other than that, really good stuff though- thanks for taking the time to work on my feedback.

andrew127:
Hi Jim,
Yes, my initial intention was to extend the ClientContext.  But after reading the code got understanding that while ClientContext  is perfect to extend for constructing single SIP transaction or for constructing dialogs,  is not so convenient to use for a sequence of independent transactions. ClientContext constructor create  the initial request, passing there headers,  etc. so for the subsequent requests (they can have different body and headers) it will be needed either to redefine the request calling new SIP.OutgoingRequest, or to do low some level processing of SIP request properties  like changing Cseq, changing headers,  etc (this is how it is done in RegisterContext).
So have taken least resistance path and just added  the ClientContext as an property redefining it for for each new request. This is not ideologically correct,  I know ( and,  there are still missed onRequestTimeout onTransportError events handlers,  they need to be added there. Will add on next update)
I see the following possible options here:
1. make the PublishContext extending the ClientContext and managing low level SIP request details, or redefining request object itself for each new transaction
2. instead of wrapping or extending  ClientContext,  add the request object as a property directly to PublishContext ( as it is done in ClientContext)

Think will not be a problem to do #1 or #2.  Please share your thoughts on this.

andrew127:
Latest commit with the PublishContext extending ClientContext.

JimGreenberg:
Hey Andrew,
looks pretty good to me- I'll let this simmer and see if anyone else in the community wants to weigh in (just in case they haven't seen in already). We're probably going to get a release together in the next few days so i'll pull it in then. Thanks again

IBlasterus:
Hi, guys.
How it use now (in version 0.20.0)?

I tried:
```
import { SIP } from 'sip.js'
import { PublishContext } from 'sip.js'
import { PublishClientContext } from 'sip.js' // https://sipjs.com/api/0.15.0/subscription/publishClientContext/
```
Always obtained error:
```
Module '"../../node_modules/sip.js/lib"' has no exported member 'SIP'.  TS2305
Module '"../../node_modules/sip.js/lib"' has no exported member 'PublishContext'.  TS2305
Module '"../../node_modules/sip.js/lib"' has no exported member 'PublishClientContext'.  TS2305
```

PR #557, opened 2018-05-15
Utils: Replace custom MD5 implementation with crypto-js/md5

### LINKED ISSUE #556: Register Forbidden after production build by webpack 4.x.x
Hi!
I update my webpack from 3.x.x up to version 4.8.1
All config is the same.

Now i get `SIP/2.0 403 Forbidden` everytime when i want to register client to server.

Server says it's because wrong password. But when i debug code in chrome dev tools the password looks ok.

I created sample project with the same problem:
https://github.com/crowmw/sip-js-test

I think it can be simillar to:
https://github.com/twilio/twilio-video.js/issues/50#issuecomment-279885506

Any help?

seanbright:
Fixes #556


seanbright:
Happy to make any changes necessary.

james-criscuolo:
I believe there is actually only one function that uses the util function, so in this case it'd be better to alter it there. Long term, we are looking to:

1) remove the Utils file
2) remove the environment philosophy, in favor of passing factories to the UA (transport is currently being moved, session description handler has already been done).

With this in mind, I would remove the changes to the environment files, as well as remove any mentions of md5 in the utils file.

Thanks,
James

seanbright:
@james-criscuolo - you're speaking a bit above my pay grade, but I took a whack at it. Let me know.

james-criscuolo:
This is much closer, but does sticking something like
```
var md5 = require('crypto-js/md5');
```
at the top of `src/DigestAuthentication.js`, then just using that, not work? Maybe it's me who is missing something here. There is still one change in the Utils file that needs to be removed as well.

Thanks again,
James

seanbright:
> Maybe it's me who is missing something here.

Nope, I'm just not that familiar/comfortable with the `import`/`require` stuff. Hopefully the most recent push addresses your concerns.

james-criscuolo:
Looks great, thanks!

PR #563, opened 2018-06-05
Fix Modifiers that strip RTP payloads

seanbright:
The first commit in this change set - while overwritten by the second - was left in as a record of a bug that was present in the previous implementation of `Modifiers.stripG722`.


seanbright:
Just bumping this to make sure it stays on your radar for the next release. Thanks.

JimGreenberg:
Yep, Egreen mentioned a small release in 'the next few days' and that turned into a little longer than that- the PR looks great so I'll just pull it in now. Thanks again for the contribution!

PR #567, opened 2018-06-12
Transport: fix loosing context for reconnect timer function

justerror:
Fixed reconnect timer error, caused by wrong context.

```
Uncaught TypeError: this.disposeWs is not a function at eval (Transport.js:343)
(anonymous) @ Transport.js:343
setTimeout (async)
Timers.(anonymous function) @ Timers.js:35
reconnect @ Transport.js:342
onClose @ Transport.js:223
```

### DIFF
--- a/src/Web/Transport.js
+++ b/src/Web/Transport.js
@@ -338,7 +338,7 @@ Transport.prototype = Object.create(SIP.Transport.prototype, {
         this.disposeWs();
         this.connect();
         this.reconnectTimer = null;
-      }, this.configuration.reconnectionTimeout * 1000);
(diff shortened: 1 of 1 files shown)


JimGreenberg:
Good catch, thanks for the PR

c960657:
Would be nice with a 0.11.1 release to get rid of this annoying bug that occurs e.g. when you have put your computer to sleep without closing your browser.

egreenmachine:
Absolutely. We are ironing out some other issues andy expect to release a new version in the next few days. Thanks again for the help.

PR #571, opened 2018-06-20
Yet another fix suggestion for stream management and media direction change on re-invites

andrew127:
Issues fixed with the patch: 

1. on re-invite setDescription() change the constraints  in unexpected manner, and this leads to wrong constraints set  on subsequent re-invites. 
To repeat: establish a call both with video and audio, change the constraints (disable video for example),  call the reinvite(). Then,  change constraints again (enable video,  disable audio) and call reinvite(). Resulting constraints on second reinvite  will be different from expected.
Proposed fix: remove constraints handling from setDescription()

2. on  re-invite, there are multiple media related artefacts (choppy video/audio, pixelated video). 
To repeat: establish a call with both video/audio, then run few subsequent re-invites disabling/enabling video. Video will become choppy,  or even fail to start (tested on latest Chrome, Win10)
Proposed fix: do the removeTrack() for each sender after the new media were acquired, but before addTtrack() was called.

### DIFF
--- a/src/Web/SessionDescriptionHandler.js
+++ b/src/Web/SessionDescriptionHandler.js
@@ -165,6 +165,21 @@ SessionDescriptionHandler.prototype = Object.create(SIP.SessionDescriptionHandle
         self.logger.error(err);
         throw err;
       })
+      .then(function removeStreams(streams) {
+        try {
+          // Remove old tracks
+          if (self.peerConnection.removeTrack) {
+            self.peerConnection.getSenders().forEach(function (sender) {
+              self.peerConnection.removeTrack(sender);
+            });
+          }
+          return streams;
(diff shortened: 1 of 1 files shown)


egreenmachine:
I see what you are saying with the constraints. I think the expected behavior here should be that if constraints are provided we use those, otherwise use the previously used constraints. Each time new constraints are passed the session descriptions handler's constraints should get overridden. The idea here being that you only need to provide constraints once per call and that all subsequent calls into the session description handler can use the same constraints unless overridden. 

I think the second change is good.

egreenmachine:
I think that this also covers #541 in a more complete manner. If we can get this into SIP.js, I would close the other PR in favor of this one.

PR #576, opened 2018-07-10
fix creating RTCSessionDescription from an RTCSessionDescription

poicooler:
Now return an instance of RTCSessionDescription instead of a plain object.

### DIFF
--- a/src/Web/SessionDescriptionHandler.js
+++ b/src/Web/SessionDescriptionHandler.js
@@ -338,10 +338,10 @@ SessionDescriptionHandler.prototype = Object.create(SIP.SessionDescriptionHandle
 
   // Creates an RTCSessionDescriptionInit from an RTCSessionDescription
   createRTCSessionDescriptionInit: {writable: true, value: function createRTCSessionDescriptionInit(RTCSessionDescription) {
-    return {
+    return new this.WebRTC.RTCSessionDescription({
       type: RTCSessionDescription.type,
       sdp: RTCSessionDescription.sdp
-    };
+    });
   }},
 
   addDefaultIceCheckingTimeout: {writable: true, value: function addDefaultIceCheckingTimeout (peerConnectionOptions) {


egreenmachine:
Thanks for the PR. However, we really do not want an `RTCSessionDescription` here. We need to be able to modify the `sdp` and according to the [documentation for `RTCSessionDescription`](https://developer.mozilla.org/en-US/docs/Web/API/RTCSessionDescription) the `sdp` property is read only. That is why I created what is essentially a `RTCSessionDescriptionInit`, the same thing as an `RTCSessionDescription` but with modifiable properties.

poicooler:
According to documentation there is also a method `toJSON` which it is not the case right now and it is causing a crash when used with [React-Native-WebRTC](https://github.com/oney/react-native-webrtc). So you might not want an `RTCSessionDescription` but the method should still be there.

poicooler:
When `this.peerconnection.setLocalDescription(sdp)` is called, the `RTCPeerConnection` of `React-Native-WebRTC` does `sdp.toJSON()` for whatever reason and then it crashes because the method is undefined due to `sdp` being your plain object. The fix without having a `RTCPeerConnection` instance would be to add the method to object.
Something like this: 
```
return {
          type: RTCSessionDescription.type,
          sdp: RTCSessionDescription.sdp,
          toJSON: () => ({sdp: RTCSessionDescription.sdp, type: RTCSessionDescription.type})
        };
```
Im not sure sure about returned values in the `toJSON` since you are telling me they are being modified. The returned values should be the updated one at all times.

egreenmachine:
According to the w3c standard:

```
The setLocalDescription method instructs the RTCPeerConnection to apply the supplied RTCSessionDescriptionInit as the local description.
```

and the spec for `RTCSessionDescriptionInit`:

```
dictionary RTCSessionDescriptionInit {
    required RTCSdpType type;
             DOMString  sdp = "";
};
```

Technically, we are conforming to spec and it sounds like the `React-Native-WebRTC` plugin is not. My recommendation here is to create your own SDH that works with the `React-Native-WebRTC` plugin. I believe that there could be other issues once we start going down the rabbit hole of getting the default SDH to work with the `React-Native-WebRTC` plugin. The goal of the default SDH is to be a very simple example that works in web browsers. There is no expectation that it will work without modification for `React-Native-WebRTC`. You can copy the default SDH and create your own for `React-Native-WebRTC` and we would gladly accept a PR back to the project for that separate SDH.

poicooler:
I'll look into it.
On the other hand, could something like this work?
```
myPeerConnection.setLocalDescription(new RTCSessionDescription(sdp));
```
I don't think this has any negative impacts.

egreenmachine:
That constructor is deprecated, which is why I would like to avoid using it. Especially in the default SDH.

> The RTCSessionDescription() constructor takes a dictionary argument, descriptionInitDict, whose content is used to initialize the new RTCSessionDescription object. This constructor is deprecated; it exists for legacy compatibility reasons only.

PR #579, opened 2018-07-12
Install jasmine-core from package.json, not just travis

jlaine:
If we list jasmine-core (which we need to run tests) in devDependencies
there is no need for a travis-specific hook. This also ensures we don't
have a version mismatch when we decide to upgrade jasmine.

Also on trusty, the default gcc version is already 4.8 so save time by
not downloading additional packages via apt.


james-criscuolo:
I believe this can actually go in the opposite direction now and we can remove jasmine for jasmine-core, then remove the dependency from travis. There used to be a peer dependency mismatch between several of our dependencies, but it looks like they've all gone jasmine-core now.

It looks like you are already changing it to the above, so my only recommendation would be to change `karma-jasmine-html-reporter` to 1.2.0 (where they updated the peer Dependency).

jlaine:
Without the extra apt downloads the build is significantly faster:

image

I will update karma-jasmine-html-reporter as requested.

jlaine:
Travis is reporting it can't find "eslint" on node 6, but it worked on previous passes. I recently saw this on another project and the fix was clear Travis's cache.

james-criscuolo:
I just did that, give it another go.

jlaine:
That did the trick, thanks!

james-criscuolo:
Thanks!

PR #580, opened 2018-07-12
Update webpack-cli to 3.x

jlaine:
Installing an old version of webpack-cli drags in babel-preset-es2015,
which gives us a warning. I was initially confused by the warning as
SIP.js itself has migrated to babel-preset-env.


james-criscuolo:
For anyone else following this, the webpack-cli jump from 2 to 3 really only affects webpack plugin developers, so this is entirely safe.

Thanks again!

PR #587, opened 2018-07-26
Reset keep-alive debounce timer

c960657:
The keep-alive debounce timer is never reset, so only one keep-alive is ever sent.


JimGreenberg:
When @egreenmachine told me about this- I didn't realize there was a PR open about it so I went and pushed up the change onto master. Thanks for the find

PR #592, opened 2018-08-15
Allow specifying second argument for RTCPeerConnection constructor

c960657:
The constructor RTCPeerConnection takes a second argument. This is deprecated (if it was ever in the spec), but AFAICT this is the only way to pass vendor-specific constrains supported by Chrome, such as `googDscp`and `googIPv6`.

In particular I am interested in enabling DSCP, but passing `googDscp` as a constraint in any other way does not seem to have any effect.

### DIFF
--- a/src/Web/SessionDescriptionHandler.js
+++ b/src/Web/SessionDescriptionHandler.js
@@ -382,6 +382,7 @@ SessionDescriptionHandler.prototype = Object.create(SIP.SessionDescriptionHandle
     options = options || {};
     options = this.addDefaultIceCheckingTimeout(options);
     options.rtcConfiguration = options.rtcConfiguration || {};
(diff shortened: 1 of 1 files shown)


egreenmachine:
Due to this not being part of the spec, and the default SDH covering the basic use cases. I am inclined to push this off to your application. You should create your own SDH copied from ours with this change integrated. Doing something like this is a prime use case for creating your own SDH.

c960657:
FWIW, I decided to monkey-patch `RTCPeerConnection` instead.

```
const realRTCPeerConnection = window.RTCPeerConnection
window.RTCPeerConnection = configuration => new realRTCPeerConnection(configuration, { optional: [{ googDscp: true }] })
```

PR #598, opened 2018-08-25
ICE gathering complete detection for old browsers

justerror:
`onicegatheringstatechange ` event implemented in chrome 59+ [MDN](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/onicegatheringstatechange#Browser_compatibility) and firefox 53+ [(bug)](https://bugzilla.mozilla.org/show_bug.cgi?id=1193731)

`triggerIceGatheringComplete` never executed in old browsers, and invite does not sended.

I made fallback with with `onicecandidate` [MDN](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/onicecandidate) : 

> If the event's candidate property is null, ICE gathering has finished.

### DIFF
--- a/src/Web/SessionDescriptionHandler.js
+++ b/src/Web/SessionDescriptionHandler.js
@@ -410,10 +410,14 @@ SessionDescriptionHandler.prototype = Object.create(SIP.SessionDescriptionHandle
       };
     }
 
-    this.peerConnection.onicecandidate = function(e) {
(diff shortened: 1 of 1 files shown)


egreenmachine:
This looks alright. I will let it sit for a few days to see if anyone else may have thoughts. Some unit tests for this case would also be nice.

PR #605, opened 2018-09-17
Initial Steps Towards TypeScript (community input requested)

james-criscuolo:
Hi,
  Internally at OnSIP we've transitioned all of our web-based projects to typescript, with a lot of success. We've been wrapping SIP.js in order to type the inputs and outputs, but believe it would be beneficial to open-source this aspect of the work, both for us and for SIP.js users.

  The initial change outlined here is minor. It is just replacing babel with typescript for our transpiling, and adding the tsconfig targeting es5. I'd be curious to hear from @jlaine here, as he introduced babel to start transpiling SIP.js, but there should be no ill effect.

One difference I did notice, the unminified dist file is 55KB larger. Upon investigating, it appears typescript forces an indent of 4 spaces (we use 2), which adds a lot of white space. They do not have a way to configure this, and explain that formatting options is a lot of added work, and as long as it's human readable it's fine, Size concerns really only come into play with minified files (The minified size has not changed). This seems like a reasonable change, and any of our users with size concerns can either minify themselves, which is expected (we point our node package to the unminified dist for this reason), or use our minified dist.

From here, there would be a fairly slow transition to full type coverage for our app. Changing all of our extensions to .ts would be a pretty quick change right after this, with no code change besides that. Once we get there, we could begin turning on and enforcing ts rules, which would provide some fairly fast benefits. We would likely allow coding to es7, and move that forward as the standard permits, with a target of es5.

Does anybody have any thoughts on this? We'll let this sit for a week or two, in case there are any concerns or questions.

Thanks,
James

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,9 +26,7 @@
     "javascript"
   ],
   "devDependencies": {
-    "babel-core": "^6.26.0",
-    "babel-loader": "^7.1.2",
-    "babel-preset-env": "^1.6.1",
+    "awesome-typescript-loader": "^5.2.1",
     "eslint": "^5.4.0",
     "jasmine-core": "^3.2.1",
     "karma": "^3.0.0",
@@ -40,6 +38,7 @@
     "karma-webpack": "^3.0.0",
     "pegjs": "^0.10.0",
     "pegjs-loader": "^0.5.4",
+    "typescript": "^3.0.3",
     "webpack": "^4.19.0",
     "webpack-cli": "^3.0.8"
   },
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,10 @@
+{
+    "compilerOptions": {
+        "outDir": "./dist",
+        "allowJs": true,
+        "target": "es5"
+    },
+    "include": [
+        "./src/**/*"
+    ]
+}
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -61,10 +61,7 @@ module.exports = function (env) {
         {
           test: /\.js$/,
           exclude: /node_modules/,
-          loader: "babel-loader",
-          options: {
-            presets: ['env']
-          }
+          loader: "awesome-typescript-loader"
         },
         {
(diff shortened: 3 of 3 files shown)


wakamoleguy:
Awesome! Full steam ahead!

jlaine:
I love typescript, but it feels as though we're mixing several concerns:

- internal code quality : we haven't even made use of the ES6 capabilities. We have ugly `.bind()` statements galore instead of using arrow functions, etc.. Updating the sip.js codebase to modern JS would be my priority.

- from an API user perspective : what I'd like most is to have typings for the sip.js API, as I usually integrate it into a codebase written in typescript. Whether or not sip.js is actually developed in typescript is irrelevant - we'll be shipping JS to the end user anyhow.

james-criscuolo:
You are absolutely correct on both points. 

- internal code quality: I would like to get rid of `.bind()` and other ugly code to move to ES6. However, there are other more pressing needs and I tend to go with the "if it is not broken, don't fix it approach". That being said, new code should definitely start to transition and eventually we will have to transition all of our code.

- Internal typing will be in the same boat as above. I'm not sure I prioritize one more highly than the other, but as we touch code, modernizing is certainly the easier of the two to accomplish.

- Typing the public-facing API would be first and foremost, as that is what the majority (perhaps all) users care about. We've already begun making the declaration files, so I would expect to release these within the next version bump or two.

I'd say another reason for this PR is to open the door to pull requests regarding typing and discussion surrounding how we go about typing, but that is likely wishful thinking.

jlaine:
I fully agree with your comment regarding move to ES6 - it's a tricky business, I had started a branch to use arrow functions in SDH but shot myself in the foot a couple of times and gave up. Let's just agree that new code always uses ES6-style, maybe consign it in a micro code style guide?

I'm looking forward to the declaration files being shipped, that's awesome. In any case - definitely no objection for me to start taking incremental steps towards typescript, great!

PR #610, opened 2018-10-02
Add diversion header to message

raphaelhovsepyan:


egreenmachine:
The idea of this is good, but I think that it needs to end up in the grammar

egreenmachine:
The `.pegjs` file is not generated, it is what should be modified.

egreenmachine:
Take a look at the commit history on the file. That should help you.

raphaelhovsepyan:
@egreenmachine @seanbright Thanks for help. Will update in few days. 🙏

james-criscuolo:
Any update here? At this point a bit more work is needed since we've moved to types, but if you still anticipate getting this done then I will leave it open.

PR #626, opened 2018-10-31
The subscription instance now emits 'terminated'

### LINKED ISSUE #395: 'terminated' event for subscriptions
If the connectivity is lost while a subscription is in `notify_wait` state, the `timer_fire()` will be invoked and, in turn `close()`. All this will happen without any event being emitted. This makes subscriptions harder to use.

Could there be a `terminated` event that is emitted whenever `this.state` is set to `'terminated'` ?

By the way, calling `subscribe()` when `sub.state === 'terminated'` should throw or emit an error. Indeed, `sub.receiveResponse` is set to no-op after calling `sub.unsubscribe()` or `sub.close()`. It is not even mentioned in the docs that this behavior is not supported.

adaxi:
Closes #395.

### DIFF
--- a/src/Subscription.js
+++ b/src/Subscription.js
@@ -13,6 +13,21 @@ SIP.Subscription = function (ua, target, event, options) {
   options = Object.create(options || Object.prototype);
   this.extraHeaders = options.extraHeaders = (options.extraHeaders || []).slice();
 
+  var state;
+  Object.defineProperty(this, 'state', {
+    get: function () {
+      return state;
+    },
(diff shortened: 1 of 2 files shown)


egreenmachine:
This is a good thing to have, but we would like it in a way that is more standardized compared to the rest of SIP.js. Can you place individual event emissions where the state is set to terminated?

egreenmachine:
Thanks. Will get this in for the next release

adaxi:
Awesome.

PR #628, opened 2018-11-11
Use arrow functions in SessionDescriptionHandler

jlaine:
Use arrow functions for promise chaining and forEach() calls instead of
having to manually bind `this` everywhere. As a result we can also
eliminate two definitions of a `self` variable.


egreenmachine:
Looks good. I have a few pending changes I am working on to get in then I will pull this in.

egreenmachine:
@jlaine looks like this has a conflict with the code I have been working on. Very sorry, but can you resolve it?

jlaine:
Sure, I'll give it a try

PR #632, opened 2018-12-18
New SDH to support React Native

mariouzae:
This new SDH is pretty much the same as standard, but with some changes to use the WebRTC implementation from React Native (https://github.com/oney/react-native-webrtc).

I create a new folder called React, but I don't know if there is the right place for it.

Tested only on Android for while. Could be able to make incoming/outcoming calls.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "sip.js",
-  "version": "0.11.6",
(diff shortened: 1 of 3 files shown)


egreenmachine:
Can you fix the lint errors and then we will take this?

mariouzae:
Done!

PR #635, opened 2018-12-28
Add some more type definitions

nud:
Here are a bunch of type definitions that we need to use SIP.js. Up until now, they were either included in `@typings/sip.js` or we were casting objects to `any`.

Here is a list of the changes:

* add constructor and `invite()` for `InviteClientContext`
    
    While it is debatable whether these are part of the public API or not,
    they are currently the only way to create a new session asynchronously,
    e.g. within a Promise.

* add missing `peerConnection` in `SessionDescriptionHandler`
    
    I set it as a non-optional field, because it is set in the constructor
    and never unset; this makes the using code easier as it removes the
    necessity to check for the `peerConnection` property to be defined.

* add missing `assertedIdentity` field in `Session`

* export `NameAddrHeader` class
    
    This is required to be able to perform such a construct:
    
    ```typescript
    import * as SIP from "sip.js"
    const h = SIP.NameAddrHeader.parse(request.getHeader("Diversion"));
    ```

* add basic type information for `Subscription`

### DIFF
--- a/types/SIP.d.ts
+++ b/types/SIP.d.ts
@@ -5,6 +5,7 @@ import { NameAddrHeader } from "./name-addr-header";
 export * from "./session";
 export * from "./session-description-handler";
 export * from "./session-description-handler-factory";
+export * from "./subscription";
 export * from "./ua";
 export * from "./uri";
 export * from "./Web"
--- a/types/SIP.d.ts
+++ b/types/SIP.d.ts
@@ -2,6 +2,7 @@ import { UA } from "./ua";
 import { IncomingMessage } from "http";
 import { NameAddrHeader } from "./name-addr-header";
 
+export * from "./name-addr-header";
 export * from "./session";
 export * from "./session-description-handler";
 export * from "./session-description-handler-factory";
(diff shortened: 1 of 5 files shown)


egreenmachine:
Just as a heads up, we are actively transitioning the entirety of the code to typescript/es6. This will hopefully make the types folder unnecessary, but at minimum the new code will match what is in the types folder at the time of the transition getting onto master.

nud:
@egreenmachine thanks for merging.

I know about these plans (it's nice); my understanding was that you would likely try to keep the interface unchanged, so the better the current interface, the more likely things won't get forgotten.

Also getting rid of `@types/sip.js` would be nice.

egreenmachine:
Yes we are trying to make sure we match these interfaces. 

Let me look into getting us removed from `DefinitelyTyped`

egreenmachine:
We have been removed from `DefinitelyTyped`. https://github.com/DefinitelyTyped/DefinitelyTyped/commit/6e2665ccc54cd52fc3b7774d3accd43e7fcc1be0

PR #638, opened 2019-01-11
Update Session.js to add mute and unmute function

anshkatriya:
Added mute and unmute for session event.
For Mute : session.mute();
For UnMute : session.unmute();

### DIFF
--- a/src/Session.js
+++ b/src/Session.js
@@ -307,6 +307,54 @@ Session.prototype = {
       }
     }
   },
+ /**
+  * Mute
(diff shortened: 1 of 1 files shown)


egreenmachine:
This commit requires internal access to the Web SessionDescriptionHandler. It will not work if you create your own SDH that does not use Peer Connection. 

Additionally, we do not want to be in the business of handling media. This is up to the users individual application.

For these reasons we will not be taking this PR.

Issue #637, opened 2019-01-08
session trackAdded event missing in type definitions


vegarda:

egreenmachine:
There is nothing here

egreenmachine:
We will get this fixed for the next release

james-criscuolo:
Fixed and on master.

Issue #230, opened 2015-09-22
Allow editing EventEmitter.defaultMaxListeners


matthieusieben:
In browsers, there is a wrapper around Node's implementation of EventEmitter. While this is very nice, the `EventEmitter.defaultMaxListeners` property is not exposed. I believe this is due to the fact that you still want to support the (deprecated) `off` method.

The following code would expose the same EventEmitter's API in browser as for Node.

``` js
Object.defineProperty(EventEmitter, 'defaultMaxListeners', {
  get: function () {
    return NodeEventEmitter.defaultMaxListeners;
  },
  set: function (value) {
    NodeEventEmitter.defaultMaxListeners = value;
  }
});
```

wakamoleguy:
> I believe this is due to the fact that you still want to support the (deprecated) off method.

Yeah.

Unless there's a strong reason, I'd rather not muck around with that code right now.  I'm hoping to finalize the deprecation (AKA, remove the wrapper) in the next major version, which shouldn't be too far off.

Is not having defaultMaxListeners affecting you adversely?

matthieusieben:
Not really, it just requires me to use `emitter.setMaxListeners(n)` instead of beeing able to use a global default.

james-criscuolo:
Closing as we no longer have an event emitter wrapper (with the upcoming version at least).

PR #639, opened 2019-01-15
Typing the Library (version 0.13.0)

james-criscuolo:
Hi Everyone,
  As we've been mentioning in some issues and google group questions, our current focus was to type the entirety of SIP.js. This PR is the result of that effort. It is not perfect typing, but it brings us a long way while attempting to preserve our public facing API. We had a hard time finding projects that have undergone a similar transition or even projects made fresh that are being used like us, so it's possible some stuff was done in a non-standard way (Enums, sigh), but the conversion was at least attempted to be done in the most by-the-book way possible.

As those who have converted a project to TypeScript know, this process has pulled out a good number of broken cases, mostly in the edge-case/very rare category. I've done a good amount of testing and plan to do more, but given the scope of the changes, I can't be certain everything is exactly as it should be. If anyone is available to test and report their results, we would appreciate the effort. **It _should_ be a drop-in replacement**, with one exception for some users (`ws_server` -> `wsServer`, see below). If it is not a drop-in replacement, we would also be curious to know what is not working, or what had to be changed.

**To test the branch**: replace your package.json dependency with our `dev-types-compiled` branch, which is this PR, with the usual release commit on top (no version change yet):

```
-    "sip.js": "0.12.0",
+    "sip.js": "onsip/SIP.js#dev-types-compiled",
```

**Known public facing API changes**:
- If you pass websocket servers as an object with a weight, `ws_server` is now `wsServer`.

**Incomplete list of potentially relevant internal changes**:
- `NameAddrHeader.parse` is now `Grammar.nameAddrHeaderParse`
- `URI.parse` is now `Grammar.URIParse`
- `SIP.EventEmitter` has been removed. At the point of removal, it was a featureless wrapper. We use the built-in one from `events` now.
- Most cases of underscored_variables have been replaced by camelCaseVariables
- Types were redone to be built into the library, instead of entirely independent. Automatic generation of declaration files does not accomplish the ability to document via types that we (eventually) want, and importing types where necessary cuts down on circular dependency possibilities.

**Known issues with the PR**
- The size of the bundles/lib folder is significantly larger. This is due to the typing of the Grammar, which we use `ts-pegjs` for. We have [an issue open](https://github.com/metadevpro/ts-pegjs/issues/29) and are hoping for a quick resolution (may submit a PR over there). **It is unlikely we will merge until this is resolved.**
- We are using the `strict` tsConfig setting, but could not turn on `noImplicitAny` because it causes the build to hang. Turns out [this is related to `ts-pegjs` as well](https://github.com/metadevpro/ts-pegjs/issues/30). It would be nice if this were fixed before merging, but I believe most of the library was already written with this turned on, so turning it on after the fact won't be bad.

**Issues to close once this is merged**
- #637: The most recent commit  adds the missing event, and a few more
- #230: We removed our EventEmitter, so this should now be doable.

**Pull Requests to close once this is merged**
- #502: Will not merge cleanly and has had no response since warning of this update coming.
- #373: Will not merge cleanly and has had no response since warning of this update coming.

Happy testing!


src/Transactions.ts
```diff
@@ -0,0 +1,834 @@
...
+  private timer_B(): void {
+    this.logger.debug("Timer B expired for INVITE client transaction " + this.id);
+    if (this.state === TransactionStatus.STATUS_CALLING) {
+      this.stateChanged(TransactionStatus.STATUS_TERMINATED);
+      this.requestSender.ua.destroyTransaction(this);
+      this.requestSender.onRequestTimeout();
+    }
+  }
+
+  private timer_D(): void {
```
seanbright: This is never called. Could be a bug or dead code, haven't looked that closely.

james-criscuolo: Yea, one thing I didn't turn on was noUnunusedLocalReferences (something to that effect). I did that for the least amount of changes possible. You'll probably see a good amount of that, I figure it can be cleaned up in a second pass, with fairly low chance of breaking anything (as its all unused).

egreenmachine: I accidentally removed this timer in an earlier commit (https://github.com/onsip/SIP.js/commit/55cb0aeff36d314e20292dbf8fb1ddbc0f3cd7b0#diff-ef9ade42e4a5eaa513b3d74c3f4f8c0cL249) by accident. It should be added back. I would like to do a full rework on the Transaction code in the future if I can find some time for it.

Issue #660, opened 2019-03-13
string log level in UA config is not working


seanplwong:
**Describe the bug**
with `new UA({ log: { level: 'debug' } })`
sipjs should log all log to console/connector

**Logs**
na

**To Reproduce (if possible)**
1. change log level to 'debug'

**Expected behavior**
log of all level should be logged to console/connector

**Observed behavior**
no log

**Environment Information**
na

**Additional context**
na

PR #661, opened 2019-03-13
normalize log level before setting

seanplwong:
Fix https://github.com/onsip/SIP.js/issues/660

### DIFF
--- a/src/LoggerFactory.ts
+++ b/src/LoggerFactory.ts
@@ -74,10 +74,10 @@ export class LoggerFactory implements LoggerFactoryDefinition {
       if (this.builtinEnabled) {
         this.print(console[Levels[levelToLog]], category, label, content);
       }
+    }
 
-      if (this.connector) {
-        this.connector(Levels[levelToLog], category, label, content);
-      }
+    if (this.connector) {
+      this.connector(Levels[levelToLog], category, label, content);
     }
   }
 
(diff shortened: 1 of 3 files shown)


james-criscuolo:
Thank you for taking a look at this, it is very close to being mergeable. I have a small issue with this line: https://github.com/onsip/SIP.js/pull/661/commits/8646cc6b072cc50481f61ad83e616dc8d7823fa9#diff-b45d4be906fc4401f1fd5133946b5250R178

We used to be able to set any string here, so if someone passed a custom connector with their own levels than I believe it would work. Now, if they were to pass that string, it would end up as undefined here.

The paths forward would either be:
1) Make the type passed in `Levels | number`
2) change the above line to set the string to what was passed if Levels[string] is undefined.

Choice 2 preserves behavior, so it should be the change made.

seanplwong:
level used to be like this:
```
    level: {
      get: function() {return level; },
      set: function(value) {
        if (value >= 0 && value <=3) {
          level = value;
        } else if (value > 3) {
          level = 3;
        } else if (levels.hasOwnProperty(value)) {
          level = levels[value];
        } else {
          logger.error('invalid "level" parameter value: '+ JSON.stringify(value));
        }
      }
    },
```
So i think custom level is not handled previously as well, it will log an error.

And even after that, it will stubble upon `LoggerFactory#genericLog()`
```
  public genericLog(levelToLog: Levels, category: string, label: string | undefined, content: any): void {
    if (this.level >= levelToLog) {
      if (this.builtinEnabled) {
        this.print(console[Levels[levelToLog]], category, label, content);
      }

      if (this.connector) {
        this.connector(Levels[levelToLog], category, label, content);
      }
    }
  }
```
when `this.level` is string, it will return false no matter what and skip everything: `'anystring' < 1000` is false.
'1' < 3 will work tho.

seanplwong:
but we could inverse the control here, just let connector to decide what it what to do.

PR #662, opened 2019-03-19
Refresh of Transactions

john-e-riordan:
Aside from a little bit of polishing with an eye toward code maintenance, the goal was to move the Transaction interface to a place where some other outstanding issues can be addressed without resorting to hacks. In particular, looking to build on this work to deal with some known race condition issues in Session that are hard to solve with the current Transaction interface (and implementation).


john-e-riordan:
With regard to breaking changes this pull request introduces, these are the changes that might or might not be considered breaking changes…

Transactions
- `checkTransaction` function has been removed
- `AckClientTransaction` class has been removed
- `InviteClientTransaction` class has a new interface (not backwards compatible)
- `InviteServerTransaction` class has a new interface (not backwards compatible)
- `NonInviteClientTransaction` class has a new interface (not backwards compatible)
- `NonInviteServerTransaction` class has a new interface (not backwards compatible)

Enums
- `TransactionStatus` has been removed

Exceptions
 - `TransportError` has a new interface (`type`, `code` and `name` properties removed)

SIPMessage
- `IncomingMessage` property `serverTransaction` renamed `transaction`

I believe that's the extent of it. The other changes to the code base were intended to be restricted to internal changes and made primarily to accommodate the changes to the Transaction interfaces. That said, a non-trivial quantity of "private" code was modified.

james-criscuolo:
After those changes, all looks well, thanks!

Issue #672, opened 2019-04-08
PRACK send with an undefined rseq value in RAck header


FThrum:
In some scenarios I see that SIP.js sends a PRACK, but the RAck header  contains "undefined" as rseq number.
The an resq does not exist in in the responses before.

Those are the relevant sip headers for an outgoing call from SIP.js 0.13.5, I removed the headers which are not relevant:
```
->
INVITE sip:123456789@null SIP/2.0
CSeq: 2641 INVITE
Allow: ACK,CANCEL,INVITE,MESSAGE,BYE,OPTIONS,INFO,NOTIFY,REFER
Supported: outbound
Contact: <sip:123@g2rohmtp9vaq.invalid;transport=wss;ob>

<-
SIP/2.0 183 Session Progress
CSeq: 2641 INVITE
Supported: replaces
Allow: INVITE, ACK, CANCEL, OPTIONS, BYE, REFER, SUBSCRIBE, NOTIFY, INFO, PUBLISH, MESSAGE
Contact: <sip:456@10.111.222.1:5060;transport=udp>

(with SDP)

<-
SIP/2.0 180 Ringing
CSeq: 2641 INVITE
Supported: replaces
Allow: INVITE, ACK, CANCEL, OPTIONS, BYE, REFER, SUBSCRIBE, NOTIFY, INFO, PUBLISH, MESSAGE
Contact: <sip:456@10.111.222.1:5060;transport=udp>

->
PRACK sip:456@10.111.222.1:5060;transport=udp SIP/2.0
CSeq: 2642 PRACK
Supported: outbound
RAck: undefined 2641 INVITE

<-
SIP/2.0 400 Bad Request
Warning: 399 sip "Syntax error in RAck"
```

**To Reproduce (if possible)**
Make a outbound call to a destination which first answers with 183 Session progress and than with 180 Ringing.
Not sure if this is timing related.
The issue does not happen if there is a 180 Ringing and then an 183 SDP (no PRACK is send as all)

**Expected behavior**
The PRACK should not be send at all, because the other side does not Indicate 100rel as Supported  or as Required.

**Observed behavior**
A PRACK is send even it the rseq is unknown

**Environment Information**
Chrome 73 on MAC OS.

**Additional context**
As a quick fix the could should check if the presense of the rseq before sending any PRACK.

In SIP.js 0.13.7:
```
22690             extraHeaders.push("RAck: " + response.getHeader("rseq") + " " + response.getHeader("cseq"));
[...]

PR #674, opened 2019-04-08
Avoid sending PRACK with undefined rseq value in RAck header

FThrum:
Before sending a PRACK it checks that the response.getHeader("rseq") exists before sending out an PRACK. I had scenarios where a PRACK was send with undefined rseq parameter in the RAck header:
RAck: undefined 2641 INVITE

Reported in #672
Could not test the code yet, I am not a javascript nor typescript developer
But feel free to change this.

### DIFF
--- a/src/Session.ts
+++ b/src/Session.ts
@@ -1664,7 +1664,7 @@ export class InviteClientContext extends Session implements ClientContext {
         return;
       }
 
-      if (response.getHeader("rseq") > 0) {
+      if (response.getHeader("rseq")) {
         extraHeaders.push("RAck: " + response.getHeader("rseq") + " " + response.getHeader("cseq"));
         this.earlyDialogs[id].pracked.push(response.getHeader("rseq"));
 
--- a/src/Session.ts
+++ b/src/Session.ts
@@ -1664,14 +1664,17 @@ export class InviteClientContext extends Session implements ClientContext {
         return;
       }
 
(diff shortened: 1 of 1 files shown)


FThrum:
Background information:
In order to send the PRACK, an Rseq number recieved in one of the 1xx provisional responses.
If the provisional 1xx responses never contains an Rseq it makes no sence to send a PRACK because the RAck header must contain an resq.
The change avoids sending an PRACK whenever there was no Rseq recieved before.

We are using this fix in so far without issues. Meanwhile I also could test the code.

egreenmachine:
Hi Fabian,

Sorry for not responding earlier. This fix kind of just covers up the underlying issue which is how SIP.js handles 100rel. We believe that the underlying issue is actually fixed with the rewrite of the core SIP.js code. We are in the process of working on this, but the first pass at it is currently on master. You can test it by building master and turning on the UA option `experimentalFeatures`. We expect to have a 0.14.0 release with this stuff finalized in the next few weeks. If the issue is still happening with the `experimentalFeatures` turned on let us know and we will take a closer look.

james-criscuolo:
This fix is now out in `0.14.0`. We ended up pulling out the `experimentalFeatures`, so you should be able to upgrade and test with no changes. Can you let us know if this is fixed?

james-criscuolo:
Closing this due to inactivity. Feel free to re-open or open a new issue if the issue persists, but we believe the issue has been resolved since `0.14.0` (suggested version is currently `0.14.2`).

PR #680, opened 2019-04-23
Emit "accepted" on BYE response

johanblumenberg:
As far as I can see, there is no event when a BYE request is completed.

From the documentation:

> ## accepted
> Fired each time a successful final (200-299) response is received.

This event was not fired on successful responses to BYE.

There is also a `terminated` and a `bye` event, but those are fired when the session is terminating, it does not wait for the BYE response.

### DIFF
--- a/src/Session.ts
+++ b/src/Session.ts
@@ -213,7 +213,9 @@ export abstract class Session extends EventEmitter {
       throw new TypeError("Invalid statusCode: " + statusCode);
     }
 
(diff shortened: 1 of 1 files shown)


egreenmachine:
I understand the issue you are trying to resolve here. However, this is going to cause a lot of confusion with our current user base. Due to that I am not going to take this. That being said there is a major refactor coming with #678 that should allow you to get a handle on this in a standard manner.

PR #708, opened 2019-06-26
issue-707

lancedouglas1:
Adding parsing support for the common header P-Asserted-Identity as a Name_Addr_Header

### DIFF
--- a/src/Parser.ts
+++ b/src/Parser.ts
@@ -80,6 +80,13 @@ export namespace Parser {
           message.fromTag = parsed.getParam("tag");
         }
         break;
+      case "p-asserted-identity":
+        message.setHeader(headerName.toLowerCase(), headerValue);
+        parsed = Grammar.parse(headerValue, "Name_Addr_Header");
+        if(parsed){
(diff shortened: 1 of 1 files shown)


james-criscuolo:
Thanks for the PR! There's still a little bit of work to do here, it looks like. If we are going to support P-Asserted-Identity, we'd like to support it in full. What this means:
- Adding a property to the `IncomingMessage` type
- Using that property within your conditional instead of calling `setHeader` again (just like all the other cases)
- Adding a property to `OutgoingRequestMessageOptions` and setting that header if the option is present.

Along with the above, in terms of your specific code changes:
- Please fix the spacing on the your conditional line (space on either side of the parentheses)
- don't use `headerName.toLowerCase()` within the case block. This isn't necessarily because it's bad, but we'd prefer to keep it in line with the other cases.

Lastly, I'm going to close the issue, we don't need to track this in both places. Thanks again!

james-criscuolo:
Hi @lancedouglas1,
  Do you have any plans to complete this work? I'll leave this open for another week or two before closing.

PR #710, opened 2019-07-08
UA: Allow omitting some logger configuration

seanbright:
I'm on the fence about making `builtinEnabled` optional, but it seems that if you've defined a `log` property, you will want at least that to be set.


james-criscuolo:
Hi Sean,
  Thanks for the PR, looks great! With the new API upcoming (see PR #705) we will likely try to break out the logger even further, but this is a good change for now, as all options should not be necessary. It will get in the next release, which we are hoping will occur within the next week or two.

Thanks again!

PR #732, opened 2019-08-13
Update migration.md

pavelsc:
according to this https://github.com/onsip/SIP.js/blob/master/docs/api/sip.js.useragentoptions.md now is `authorizationUsername` and `authorizationPassword` not `authenticationUsername` and `authenticationPassword`


james-criscuolo:
Good catch, thanks!

Issue #738, opened 2019-08-29
Unspecified transport error


Nitrino:
**Describe the bug**
I tried to make REGISTER request and get an errors before the request is sent.
I follow the examples from the documentation, maybe I need to specify some additional connection parameters?

```
sip.user-agent-client | Unspecified transport error.
sip.user-agent-client | User agent client request transport error. Generating internal 503 Service Unavailable.
sip.transaction.nict | Unspecified transport error.
sip.transaction.nict | Transport error occurred in non-INVITE client transaction with id z9hG4bK8365863.
sip.transaction.nict | Failed to send initial outgoing request.
```

**To Reproduce (if possible)**
Code: 
```typescript
import {
  UserAgent,
  UserAgentOptions,
  RegistererOptions,
  Registerer
} from "sip.js/lib/api"

const uri = UserAgent.makeURI("sip:alice@localhost");
const userAgentOptions: UserAgentOptions = {
  uri: uri,
  authorizationUsername: "user",
  authorizationPassword: "password",
  transportOptions: {
    wsServers: [{ wsUri: "ws://localhost:4000/sip/websocket" }]
  },
};

const userAgent = new UserAgent(userAgentOptions);
userAgent.start().then(() => {
  const registererOptions: RegistererOptions = {};
  const registerer = new Registerer(userAgent, registererOptions);
  registerer.register();
});
```

**Logs**
```
Fri Aug 30 2019 00:30:30 GMT+0300 (Moscow Standard Time) | sip.UserAgent | Configuration:
Fri Aug 30 2019 00:30:30 GMT+0300 (Moscow Standard Time) | sip.UserAgent | · allowLegacyNotifications: false
Fri Aug 30 2019 00:30:30 GMT+0300 (Moscow Standard Time) | sip.UserAgent | · allowOutOfDialogRefers: false
Fri Aug 30 2019 00:30:30 GMT+0300 (Moscow Standard Time) | sip.UserAgent | · authorizationPassword: NOT SHOWN
Fri Aug 30 2019 00:30:30 GMT+0300 (Moscow Standard Time) | sip.UserAgent | · authorizationUsername: "user"
Fri Aug 30 2019 00:30:30 GMT+0300 (Moscow Standard Time) | sip.UserAgent | · autoStart: true
Fri Aug 30 2019 00:30:30 GMT+0300 (Moscow Standard Time) | sip.UserAgent | · autoStop: true
[...]

PR #739, opened 2019-09-03
Doc: call register after transport connected

Nitrino:
Improve docs related with #738 issue


james-criscuolo:
Closing in favor of a commit in the other PR: https://github.com/onsip/SIP.js/pull/741/commits/056c09d663a9e8f2a7428dad68d1a7beb13bb4f3

The issue is that `autostart` leads to you writing more complicated code, so we are deprecating that. Defaulting it to `false` allows the README to remain as is.

PR #740, opened 2019-09-04
Fix type of SessionDescriptionHandlerOptions.constraints

nud:
Elsewhere in the Web SessionDescriptionHandler we assume the constraints property to be of type MediaStreamConstraints, but for historical reasons the SessionDescriptionHandlerOptions property had a custom interface type.

This made it difficult to cast options because the Typescript compiler would complain on such construct:

```
  const constraints: MediaStreamConstraints = {
    audio: true,
    video: false,
  }

  const options: SessionDescriptionHandlerOptions = { constraints }
```

The raised error was:

```
  ERROR TS2345: Argument of type '{ sessionDescriptionHandlerOptions:{ constraints: MediaStreamConstraints; }; }' is not assignable to parameter of type 'Options'.
  Types of property 'sessionDescriptionHandlerOptions' are incompatible.
    Type '{ constraints: MediaStreamConstraints; }' is not assignable to type 'SessionDescriptionHandlerOptions'.
      Types of property 'constraints' are incompatible.
        Type 'MediaStreamConstraints' is not assignable to type '{ audio: boolean; video: boolean; }'.
          Types of property 'audio' are incompatible.
            Type 'boolean | MediaTrackConstraints | undefined' is not assignable to type 'boolean'.
              Type 'undefined' is not assignable to type 'boolean'.
```

Since `MediaStreamConstraints` is more permissive than the previous type `{ audio: boolean, video: boolean }`, it should not break existing code.

### DIFF
--- a/src/session-description-handler.ts
+++ b/src/session-description-handler.ts
@@ -76,7 +76,7 @@ export type SessionDescriptionHandlerModifiers = Array<SessionDescriptionHandler
  */
 export interface SessionDescriptionHandlerOptions {
   modifiers?: SessionDescriptionHandlerModifiers;
-  constraints?: { audio: boolean, video: boolean };
+  constraints?: MediaStreamConstraints;
 }
 
 /**


john-e-riordan:
I took a look at using a generic type and it opens a bit of a can of worms as the generic type is apparently hard to isolate to the SDH. I agree that as `constraints` are simply passed blindly along to the SDH (sip.js does not use them in any way), it makes sense to allow for a more permissive type. For now, gonna propose...
```
interface SessionDescriptionHandlerOptions {
  constraints?: object;
}
```
which addresses the complier error associated with...
```
  const constraints: MediaStreamConstraints = {
    audio: true,
    video: false,
  }

  const options: SessionDescriptionHandlerOptions = { constraints }
```
And since `object` is more permissive than the previous type `{ audio: boolean, video: boolean }`, it should not break existing code.

PR #754, opened 2019-11-11
Fixing issue when sending 0 as integer in dtmf.

PerezArmando:
I was getting the following error when sending 0 as an integer to the session's dtmf function.

image

Steps to reproduce:
1. Open a voice session, and call session.dtmf(0).
2. Open the browser's console and lookup for the error.

Tested on:
1. Google Chrome Version 78.0.3904.97 (Official Build) (64-bit)
2. Angular 8.2.7
3. SIP.js 0.14.0 (even though, the up to date version contains the same issue).

### DIFF
--- a/src/Session/DTMFValidator.ts
+++ b/src/Session/DTMFValidator.ts
@@ -17,6 +17,6 @@ export class DTMFValidator {
   }
 
   private static generateInvalidToneError(tone: string): void {
-    throw new TypeError("Invalid tone(s): " + tone.toLowerCase());
+    throw new TypeError("Invalid tone(s): " + !!tone ? tone.toLowerCase() : tone);
   }
 }
--- a/src/Session/DTMFValidator.ts
+++ b/src/Session/DTMFValidator.ts
@@ -17,6 +17,6 @@ export class DTMFValidator {
   }
 
   private static generateInvalidToneError(tone: string): void {
-    throw new TypeError("Invalid tone(s): " + !!tone ? tone.toLowerCase() : tone);
+    throw new TypeError("Invalid tone(s): " + (!!tone ? tone.toLowerCase() : tone));
   }
 }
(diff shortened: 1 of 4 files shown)


seanbright:
There is similar code [here](https://github.com/onsip/SIP.js/blob/388d1b84e7e806d89ece3a76175b9e41890ad048/src/Session/DTMF.ts#L39-L53):

https://github.com/onsip/SIP.js/blob/388d1b84e7e806d89ece3a76175b9e41890ad048/src/Session/DTMF.ts#L39-L53

Maybe this should be extracted out somehow so it could be used from both places?

PerezArmando:
Thanks for pointing out that @seanbright. I will work on that and update the PR.

src/Session/DTMFValidator.ts
```diff
@@ -0,0 +1,17 @@
...
+    // Check tone type
+    if (typeof tone === "string" ) {
+      tone = tone.toUpperCase();
+    } else if (typeof tone === "number") {
+      tone = tone.toString();
+    } else {
+      throw new TypeError("Invalid tone: " + tone);
+    }
+    // Check tone value
+    if (!tone.match(/^[0-9A-D#*]$/)) {
```
PerezArmando: Show we allow the comma here?

seanbright:
IMO, yes, but I am not one of the project maintainers so I can't say if that is acceptable. It does seem to be supported by [`RTCDTMFSender.insertDTMF()`](https://developer.mozilla.org/en-US/docs/Web/API/RTCDTMFSender/insertDTMF).

PerezArmando:
@seanbright, I ran into some problems running some npm scripts on my windows machine, so I had to commit in order to check every time that I did not break any tests. Are you working on Windows?

egreenmachine:
This looks great and I am going to take it. Just be aware that this is all on our "legacy" API and will not be taken forward in our 0.16 release.

PR #774, opened 2020-02-21
Fix reinvite when using authorizationUser

jahndis:
Currently hold/unhold are not working when using a SIP backend that requires an authorizationUser in the UA configuration. When the reinvite gets sent, it does not have the authorization header set, so the 401 auth flow never happens.

This fix adds the authenticationGuard to the re-invite-user-agent-client receiveMessage method (like the regular invite). Which appears to fix the hold issue.
However, after adding this, unhold failed to work because the cseq was not incrementing correctly. It appeared this was due to the "options.cseq" being passed in to dialog.ts createOutgoingRequestMessage() which did not set the internal localSequenceNumber state, resulting in the same cseq being used for successive messages.

This fix also adds logic to set the localSequenceNumber to the passed in options.cseq which appears to increment it properly.
NOTE: For some reason this second change is causing several unit tests to fail with:
```
Error: Jasmine Clock was unable to install over custom global timer functions. Is the clock already installed?
```
for reasons that are not very clear to me.

Any thoughts on whether this is the appropriate fix for reinvite when using an authorizationUser are welcome, but so far, these couple of changes are working in our setup.

Final note: We are currently using 0.14.6 (waiting for API to stabilize in 0.16), so if this fix gets merged, we'd want to backport it to that version if possible.

Thanks in advance for your feedback.

### DIFF
--- a/src/core/dialogs/dialog.ts
+++ b/src/core/dialogs/dialog.ts
@@ -533,6 +533,16 @@ export class Dialog {
   }
 
   /**
+   * Used to synchronize the local sequence number of a dialog with an
+   * incoming response message. (Used in the case of a re-invite that
+   * goes through 401 authentication).
+   * @param message - Incoming response message to sync local sequence number.
+   */
+  public updateDialogSequenceNumber(message: IncomingResponseMessage): void {
+    this.dialogState.localSequenceNumber = message.cseq;
+  }
+
+  /**
    * If the remote sequence number was not empty, but the sequence number
    * of the request is lower than the remote sequence number, the request
    * is out of order and MUST be rejected with a 500 (Server Internal
@@ -576,15 +586,5 @@ export class Dialog {
       return false;
     }
     return true;
-  }
-
-  /**
-   * Used to synchronize the local sequence number of a dialog with an
-   * incoming response message. (Used in the case of a re-invite that
-   * goes through 401 authentication).
-   * @param message - Incoming response message to sync local sequence number.
-   */
-  public updateDialogSequenceNumber(message: IncomingResponseMessage): void {
-    this.dialogState.localSequenceNumber = message.cseq;
   }
 }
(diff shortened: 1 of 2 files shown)


egreenmachine:
The PR looks good except for the failing tests. I think that the first failure is a "true" failure. So accept after a reliable progress is not working properly anymore. I think that fixing that will fix the timer failing to install.

jahndis:
Took some time to dig into the failing tests. It appeared my initial change of synchronizing the local sequence number in createOutgoingRequestMessage() was a bit too heavy handed, and the cseq was being incorrectly set when a BYE was called after a PRACK (during the tests' afterEach()).

I've removed that change, and added a function that updates the local sequence number on a dialog, and I call that in the re-invite receiveResponse when it's handling a 200, and that appears to be working
 (and should have much less functional impact).

I've tested hold/unhold against backends with and without authorization and both appear to be working as far as I can tell.

Let me know if there's anything else you'd have me change.

jahndis:
Has anyone had a chance to look at this further?

john-e-riordan:
Hey,

Sorry about the delay. I've pulled this into a branch of mine, modified it a bit, and added a test.

While your implementation works, I wanted to cover the general case of any in dialog requests being authenticated, not just re-INVITE. I also wanted to update the local sequence number based on what the dialog knows it to currently be instead of depending on the response message which seems a bit fragile (it assumes the CSeq is the response is the most current which may not be true and it also assumes the remote peer is behaving correctly).

I should have a pull request up tomorrow which includes all that and I believe that will get onto master and thus into the next 0.15.x release which I understand is coming soonish.

james-criscuolo:
Hi @jahndis,
  Can you test or take a look at #787 and let us know if it satisfies what you are looking for? If it does, we'll close this PR.

jahndis:
@james-criscuolo I pulled the relevant changes into the 0.14.6 branch we are using, and it all appears to work fine. Feel free to close this PR.

Is there a possibility of getting the changes you made backported to the 0.14 release branch?

Thanks for taking a closer look!

james-criscuolo:
Thanks for testing, I'll close this PR. In terms of backporting, 0.15.x is a drop replacement for 0.14.x, so you should have no issues updating, and I believe the only changes to the old API are non-breaking bug fixes, so you are likely better off updating anyway.

PR #784, opened 2020-04-10
Handle MESSAGE in dialog

manuquentin:
This PR aims to be able to send sip `MESSAGE` inside a current dialog.
- [x] Responds to `MESSAGE` with 200 OK
- [x] Add an `onMessage` event on the session when a message is received
- [x] Add a `message` method in SessionDialog to be able to send MESSAGE.

We've used jsSIP in a POC that allows to send this kind of message inside a dialog, and we don't found a way to do it with `Messager`.

### DIFF
--- a/src/Session.ts
+++ b/src/Session.ts
@@ -16,6 +16,7 @@ import {
   IncomingRequestMessage,
   IncomingResponse,
   IncomingResponseMessage,
+  IncomingMessageRequest,
   InviteServerTransaction,
   Logger,
   NameAddrHeader,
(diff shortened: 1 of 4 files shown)


john-e-riordan:
I pull this into a branch of mine and on top of it I did...
- lint
- a bit of cleanup
- added a tests
- ported it to the new API

The one functional change I made in a nod to consistency was in the old API was the event it emits is "message" which is the same as the out of dialog case (instead of "onMessage" as in the pr). As the old API event names are not known for their consistency I'm not sure this was worth doing but I couldn't help myself.

john-e-riordan:
Hi @manuquentin,

If possible could you test or look at #787 and see if it satisfies your needs?

Thanks

sboily:
Hello @john-e-riordan, for me looks like good, thank you!

james-criscuolo:
I'm going to close this, as it's covered in #787. Thanks again!

PR #798, opened 2020-04-30
Fix ACK re-INVITE race when offer in 200 Ok

john-e-riordan:
There is a known race condition where the callee receives re-INVITE (established state) while in the moratorium state (waiting for ACK) which we currently throw an error on (we crash). The race is documented here: https://tools.ietf.org/html/rfc5407#section-3.1.5

It is not something that comes up if the application layer waits for a short period of time (like 2 seconds) before sending a re-INVITE after an ACK as per implementation recommended best practice, but we shouldn't crash regardless.

This fixes it so we respond with a 500 response per the specification.

@james-criscuolo brought this issue to my attention on 1/14.

### DIFF
--- a/docs/core/sip.js.sessiondelegate.onack.md
+++ b/docs/core/sip.js.sessiondelegate.onack.md
@@ -9,7 +9,7 @@ Receive ACK request.
 <b>Signature:</b>
 
 ```typescript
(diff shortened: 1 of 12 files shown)


john-e-riordan:
I updated this so that it is backward compatible.

PR #806, opened 2020-05-06
Migrate from TSLint to ESLint + Prettier

john-e-riordan:


seanbright:
I am sure it is no longer a problem, but [some context](https://github.com/onsip/SIP.js/issues/556) on why the project was using the `crypto-js` MD5 implementation.

james-criscuolo:
That issue came up in several different instances, all surrounding using a minified bundle, then minifying it further. At the time, our default path was to include our minified bundle, Given that our default path provides unminified (and unbundled) code, I don't think it should be an issue any longer.

The bundles in `dist` are still provided via NPM, including the minified one, but we do plan on removing those from our npm package at some point in the near future.

PR #820, opened 2020-06-07
Add ha1 authentication option to digest authentication

grxnola:
Adds a method of authorising a user with a service by providing SIP.js with an authorisation token instead of password, which marginally helps security-wise. I haven't written any tests because I don't really know how, but I did integrate the ha1 into the exising mock code.


john-e-riordan:
This look ok to me as is. The documentation should be updated (`npm run build-docs`).

PR #871, opened 2020-10-26
re-REGISTER on 50% of expires time

qermit:


egreenmachine:
We do not think this is necessarily the right decision on when to register. It may be good for your environment, but not everyone else's environment. Our solution of registering 3 seconds before expiration suffers the same downfall. The best thing to do here would be to make it configurable via `registrationOptions`.  If you make it configurable we would accept it, but otherwise we are going to reject this and will try and make it configurable in a future release. Thanks.

Issue #872, opened 2020-11-04
Missing '#' in grammar marks


er1cAk:
**Is your feature request related to a problem? Please describe.**
In some cases I need use '#' in Inviter but symbol '#' is not allowed in grammar marks and I can't make URI with '#'

**Describe the solution you'd like**
Append '#' to [mark's](https://github.com/onsip/SIP.js/blob/master/src/grammar/pegjs/src/grammar.pegjs#L27)

seanbright:
That's not valid according to [RFC 3261](https://tools.ietf.org/html/rfc3261#section-25.1):

    mark =  "-" / "_" / "." / "!" / "~" / "*" / "'" / "(" / ")"

If you want a `#` in a URI, you will need to escape it as `%23`

er1cAk:
thank you

PR #873, opened 2020-11-04
Update of grammar marks

er1cAk:
PR for #872


seanbright:
That's not valid according to [RFC 3261](https://tools.ietf.org/html/rfc3261#section-25.1):

    mark =  "-" / "_" / "." / "!" / "~" / "*" / "'" / "(" / ")"

If you want a `#` in a URI, you will need to escape it as `%23`

Issue #870, opened 2020-10-26
shorter or flexlible re-REGISTER time


qermit:
https://github.com/onsip/SIP.js/blob/f41b9c9aac39fce991b47b439f8f13e708d7fbf3/src/api/registerer.ts#L659

The re-REGISTER be performed before previous one expires. 3 seconds is perfect in case of perfect connection but might be not long enough in real world.

I suggest to use either 50% of **expires** time or configurable.

qermit:
My PR
https://github.com/onsip/SIP.js/pull/871

james-criscuolo:
Released in 0.20.0.

PR #902, opened 2021-02-05
Make the interval at which re-REGISTER requests are sent configurable

erogleva:
as suggested in #870

### DIFF
--- a/src/api/registerer-options.ts
+++ b/src/api/registerer-options.ts
@@ -46,4 +46,10 @@ export interface RegistererOptions {
    * @defaultValue domain portion of the user agent's uri
    */
   registrar?: URI;
+
+  /**
+   * Determines when a re-REGISTER request is sent. The value should be specified as a percentage of the expiration time (between 50 and 99).
+   * @defaultValue 99
+   */
+  refreshFrequency?: number;
 }
--- a/src/api/registerer.ts
+++ b/src/api/registerer.ts
@@ -22,10 +22,12 @@ import { UserAgent } from "./user-agent";
  */
 export class Registerer {
   private static readonly defaultExpires = 600;
+  private static readonly defaultRefreshFrequency = 99;
 
   private disposed = false;
   private id: string;
   private expires: number;
+  private refreshFrequency: number;
   private logger: Logger;
   private options: RegistererOptions;
   private request: OutgoingRequestMessage;
@@ -125,6 +127,14 @@ export class Registerer {
       throw new Error("Invalid expires.");
     }
 
+    // Interval at which re-REGISTER requests are sent
+    this.refreshFrequency = this.options.refreshFrequency || Registerer.defaultRefreshFrequency;
+    if (this.refreshFrequency < 50 || this.refreshFrequency > 99) {
+      throw new Error(
+        "Invalid refresh frequency. The value represents a percentage of the expiration time and should be between 50 and 99."
+      );
+    }
+
     // initialize logger
     this.logger = userAgent.getLogger("sip.Registerer");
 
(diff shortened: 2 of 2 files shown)


james-criscuolo:
This Pull Request changes the code to remove the use of registrar's response containing the actual expires time. The code now assumes that the expire time passed in (or the default) is the actual expire time, which is not always the case. The registrar has the ability to set the time lower than what was requested. Making this configuration option be a percentage of the actual expires (so 1-99) would cover this.

On a related note, this change also allows the user to set a reRegisterInterval that will get them in trouble if they set it too high. The set interval should either be ignored or dismissed with an error message if this occurs. The percentage solution would also squash this.

I recognize I've suggested the percentage solution twice, but I'm sure other options could also be valid.

erogleva:
I see. Perhaps the configuration option could then also be the number of seconds by which the `expires` time is decreased (so that it replaces the value which is currently 3 seconds). Though the percentage solution seems better to me for the case when the registrar sets the time lower than what is requested.

erogleva:
I've reworked the code so that the value passed as an option represents a percentage of the actual expiration time

james-criscuolo:
This looks much better, one small change. We want this to be named refreshFrequency, this name aligns more with RFC 3261. We will get around to taking the PR as is and making the changes on top, but if you want to make these changes prior to that, that would be great.

Thanks again!

seanbright:
Just some disinterested party thoughts - 1% seems like a really low lower bound. 50% seems more reasonable to me.

james-criscuolo:
We also had that thought and shrugged it off, but you're right, we should keep the lane as safe as possible for the unknowing user out there. That's another change to get in if the PR isn't changed by the time we take it, thanks.

src/api/registerer-options.ts
```diff
@@ -46,4 +46,10 @@ export interface RegistererOptions {
    * @defaultValue domain portion of the user agent's uri
    */
   registrar?: URI;
+
+  /**
+   * Determines when a re-REGISTER request is sent. The value should be specified as a percentage of the expiration time (between 1 and 50).
```
seanbright, line 51: (between 50 and 99)

src/api/registerer.ts
```diff
@@ -22,10 +22,12 @@ import { UserAgent } from "./user-agent";
  */
 export class Registerer {
   private static readonly defaultExpires = 600;
+  private static readonly defaultRefreshFrequency = 50;
```
seanbright, line 25: Maybe default to 99? For a 1 hour expiration this would be about 36 seconds before expiration.

src/api/registerer.ts
```diff
@@ -125,6 +127,14 @@ export class Registerer {
       throw new Error("Invalid expires.");
     }
 
+    // Interval at which re-REGISTER requests are sent
+    this.refreshFrequency = this.options.refreshFrequency || Registerer.defaultRefreshFrequency;
+    if (this.refreshFrequency < 1 || this.refreshFrequency > 50) {
```
seanbright, line 132: 50 and 99

src/api/registerer.ts
```diff
@@ -125,6 +127,14 @@ export class Registerer {
...
+      throw new Error(
+        "Invalid refresh frequency. The value represents a percentage of the expiration time and should be between 1 and 50."
```
seanbright, line 134: 50 and 99

src/api/registerer-options.ts
```diff
@@ -46,4 +46,10 @@ export interface RegistererOptions {
...
+   * @defaultValue 50
```
seanbright, line 52: Maybe 99?

erogleva:
Now I see I had completely misunderstood your comment :sweat_smile: It should be fixed now. As for the default value, 99 is the closest to the current behavior... or otherwise values like 99.9 should also be allowed.

james-criscuolo:
This looks great now, thanks! And thank you @seanbright for the reviewing assist!

PR #904, opened 2021-02-05
strip TCP candidates with unix line ending

robsonhermes:
When using sip.js on linux computers, the ice candidates have unix line endings. That is,  \n only.

This change targets fixing the regular expression to match these cases also.

Sorry if I missed something, this is my first PR here.


james-criscuolo:
If you run `npm run lint` you'll see a failure due to the formatting of the change. I think `npm run lint-fix` will fix, otherwise you will need to address manually.

robsonhermes:
hello @james-criscuolo 
Thanks for that. Sorry, I'm a Java developer. Jumping in typescript for the first time.

Using lint-fix did the trick.

robsonhermes:
Sorry, my first comment could have been better.
There is a method inside the api to strip TCP ice candidates. The api does it by using a regular expression to match those. However, this reg exp contains a line ending and it assumes line ending as `\r\n`. If we are running sip.js in Unix-like OS, then these TCP candidates will not be stripped away.
The change aims to match both line endings.

seanbright:
Out of curiosity, when you say:

> When using sip.js on linux computers

How is sip.js being run? In a browser or via node or... ? I ask because the SDP RFC says that lines are `\r\n` terminated so I am wondering where the invalid SDP is coming from.

robsonhermes:
Hello @seanbright 

We are running sip.js inside an Electron app. So technically both are true: in browser AND via node. It's a nodejs project, but Electron provides a Chromium instance.

Anyway, I got the information from a colleague who was doing the troubleshooting, but now I realized that it was never set in the `Modifiers`, the `stripTCPCandidates` modifier. Sorry about the confusion.

Will close this PR.

Issue #909, opened 2021-02-20
Allow extraHeaders for invitation.accept()


danvirsen:
**Is your feature request related to a problem? Please describe.**
Yes, I'm currently upgrading from 0.11 to 0.19 and this is something that we used in that version.
We have proprietary headers that are sent with an INVITE from a legacy client. The client that uses SIP.js needs to answer with matching headers if the headers were part of an incoming INVITE.

**Describe the solution you'd like**
Add support to supply extra headers when accepting an incoming invite.
Both `InvitationProgressOptions` and `InvitationRejectOptions` contain extraHeaders, but not `InvitationAcceptOptions`.
Is there a reason why the option to add extra headers to the 200 Ok on an incoming invite was removed? I can't find anything in any changelogs (although I can't say I've read every commit message, so I might have missed it).

**Describe alternatives you've considered**
I posted [a question in the Google Group](https://groups.google.com/g/sip_js/c/32hPAKelwWQ) and got [an answer with a suggestion](https://groups.google.com/g/sip_js/c/32hPAKelwWQ/m/lezWBeYmBAAJ), but sadly that workaround won't work since the property that I would need to override (`Invitation.autoSendAnInitialProvisionalResponse`) is readonly.
I could simply trigger another `Invitation.progress()` with the extra headers, but that still wouldn't help in my case since the other clients, which I have no control over, only checks the headers on the 200 Ok.
Even if it would work, it seems pretty hacky.

Edit: It's possible to override `sendInitialProvisionalResponse` on the `UserAgent` instead, and manually trigger a progress on invitations with the extra headers. But as I mentioned before it still doesn't solve my issue since I need the headers on the 200 Ok instead.

danvirsen:
Thanks for the quick reply @james-criscuolo !
I'll close this issue.

PR #910, opened 2021-02-22
Add support for extraHeaders on Invitation.accept

danvirsen:
Added extraHeaders to the Invitation.accept method to fix issue https://github.com/onsip/SIP.js/issues/909.

I'm not familiar with the repository, but I think the implementation looks similar to progress and reject.


james-criscuolo:
This looks good, thank you. I spoke to John about this and it looks like it's an inconsistency that we can't explain. We hope to do a pass over the whole library to make sure extraHeaders are consistent, but this is a needed addition for now.

PR #911, opened 2021-02-28
Add `onAck` delegation for Session

slavikbialik:
Added the ability to make an `onAck` session delegation, so we'll be able to catch when we are getting a SIP ACK from PBX side.
This helps to reduce race conditions, because currently the only way to detect when a call is REALLY being answered is to listen on the session state when it becomes `SessionState.ESTABLISHED`, usually it becomes ESTABLISHED before the ACK arrives. But waiting for ACK and acting accordingly is more reliable way.

Note: this is my first PR here, I hope it is done according to your guidelines (which I couldn't find).


john-e-riordan:
* A request to end a {@link Session} (incoming ACK).

slavikbialik:
@john-e-riordan for some reason the CI failed as a result of a single test failure. Is there a way to make it re-run the tests without making any additional commit? Because the failed test should not fail because I just changed a comment 😓

james-criscuolo:
Thanks again! Not sure what's happening with Travis, we saw it elsewhere too, will have to investigate.

PR #928, opened 2021-05-10
Added ready() method on ICE candidate event

dixoncb:
This modification to the RTCPeerConnectionIceEvent notfication allows the client to inspect the incoming event and, if appropriate, to terminate ICE gathering and proceed immediately to SDP generation with the current collection of ICE candidates. It can reduce connection time significantly, for example, when simply an srflx or relay candidate is needed for a client with multiple IP addresses. The code intercepts the RTCPeerConnectionIceEvent and wraps it in an object of type RTCPeerConnectionIceEventEx, which presents a a ready() method. When the latter is called, iceGatheringComplete() is called on the SessionDescriptionHandler instance.

In my use case - connecting to a Kamailio proxy - call setup time is reduced from about 8 to about 2 seconds. 

An alternative approach would be to deep-clone the RTCPeerConnectionIceEvent and add the ready() method, but this preserves the original WebRTC specification for the event.


egreenmachine:
I am leaning towards rejecting this. We provide functionality to override the default SDH and this seems like the perfect use case to override the default SDH to achieve some custom logic for your application. This is not to WebRTC standard as far as I can tell.

dixoncb:
Thanks. That works. It's a bit indirect, but if I give the SessionDelegate a method which provides me with a reference to the SessionDescriptionHandler, I can call the iceGatheringComplete() method on that reference. Trouble is, I have to cast the SessionDescriptionHandler as an any, because that method isn't published, which makes me uneasy. Is there another way? 

I'm basically trying to replicate the action produced by iceGatheringTimeout. I agree that this is not part of the WebRTC standard (I'm not sure that iceGatheringTimeout is either, though), but it does provide a way to launch the connection the instant we know we have a suitable route. It really does produce a fast connection in what I suspect is a common use case - trying to route an SRTP connection to a public server.

egreenmachine:
The `iceCheckingTimeout` is definitely not part of the WebRTC spec.

You can copy our SDH and modify it. Then create a factory for it and pass that into the UA [sessionDescriptionHandlerFactory](https://github.com/onsip/SIP.js/blob/master/docs/api/sip.js.useragentoptions.sessiondescriptionhandlerfactory.md). 

There is some outstanding work on our end to make it easier to extend our SDH - but we have not gotten to it yet. So this is probably the best bet for now.

dixoncb:
OK. That's what I'll do. I'll keep checking back, because I'd rather avoid cutting and pasting swathes of the project. I'd ask you to look at this feature though. I'd like to claim it was my idea, but it is implemented in another WebRTC project, and it works well in reducing connection times in internet-routed scenarios. That said, you've done a superb job on this project.  I need to fit a tricky orthogonal state model on top of a websocket SIP connection, and only sip.js gives me the close level of control I require. 

I'm happy for you to close this one and reject the pull request. Thanks for your attention.

PR #938, opened 2021-06-29
Add earlyMedia support to SimpleUser object

7i77an:
Set earlyMedia parameter and setup remote and local media if apply into Established SessionState.

		this.m_userAgent.call(`sip:${number}@${this.rtmpServer}`, {
			extraHeaders : [
				...this.m_defaultHeaders
			],
			earlyMedia : true
		})

I need to condition the remote media play because when closing the session it generated an error:
The play () request was interrupted by a new load request.

I did not test video but audio works correctly in the different scenarios.

### DIFF
--- a/src/platform/web/simple-user/simple-user.ts
+++ b/src/platform/web/simple-user/simple-user.ts
@@ -1011,10 +1011,12 @@ export class SimpleUser {
       }
       mediaElement.srcObject = localStream;
       mediaElement.volume = 0;
-      mediaElement.play().catch((error: Error) => {
-        this.logger.error(`[${this.id}] Failed to play local media`);
-        this.logger.error(error.message);
-      });
+      mediaElement.onloadedmetadata= function() {
+        mediaElement.play().catch((error: Error) => {
+          this.logger.error(`[${this.id}] Failed to play local media`);
+          this.logger.error(error.message);
+        });
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


egreenmachine:
Can you clean this up and provide some instructions and documentation on how it works?

john-e-riordan:
I understand you are looking for SimpleUser to support early media. And I see how you are trying to make it happen. A couple of comments...

1. Early media is only about playing the remote media (incoming media), so there is no need to setup local media. 
2. Trying to use the Establishing state of the Session is not the best way to do it. The Establishing state is reached before there is any media available to play (and there may never be any media to play). The recommended way to get a handle on early media is to instead setup an `onProgress` request delegate for the outgoing INVITE which can look for 183 Session Progress responses and setup media when/if one arrives. If the `earlyMedia` option was set when making the call (see below), the remote media will be available when `onProgress` is called.

I've coded it up and tested it. It makes for a good example of how to best handle early media which is a question that comes up. Use it like you were expecting to be able to...

```
  simpleUser
    .call(target, {
      earlyMedia: true
    })
```

It's on the [main branch](https://github.com/onsip/SIP.js/tree/main) in 5b0c29ebd00c7e678b9bd2616eb5e8aef9e483d1 and will get out in the next release.

It's also worth noting that using early media requires that the INVITE request MUST NOT fork. And attempting to utilize early media will fail if `inviteWithoutSdp` is true. (See the docs for more info).

PR #940, opened 2021-07-08
fix typo in docs/session-description-handler.md

nnaoi:
I always use this library. Thank you for the great library.


onsip:
Thank you!

PR #948, opened 2021-08-25
Add `onCancel` delegate for Session

slavikbialik:
This PR will add `onCancel` delegation for Session. This can help developers to detect CANCEL events while call is ringing on the client and fire any action needed. It can assist detect 2 scenarios:

1. CANCEL arriving before answering the call.
2. CANCEL arriving while trying to answer the call and Session state is still `ESTABLISHING` (means 200OK - final response - has not yet sent out).

This PR is not breaking the RFC and SIP.js internal mechanism to treat incoming CANCEL requests.


john-e-riordan:
Thanks for the pull request and thanks for reworking per Eric's suggestions!

I cleaned up the comments and added some tests for `onCancel` as well as the earlier `onAck` contribution.

Pulled in the unrelated fix for the `unsubscribe` timer N issue as well.

It's on the [main branch](https://github.com/onsip/SIP.js/tree/main) and will go out in next release.

wnecloud:
@slavikbialik,

Hi, We are using SIP.JS 0.20.1 version and I'm using the onCancel functionality.  Problem is, the onCancel is triggered for real cancelled calls by caller and also for picked up calls elsewhere. No it is still unclear for me to filter out or catch if the call has been answered elsewhere or being really cancelled.

On browser, I'm seeing PBX gives through Reason: SIP;cause=200;text="Call completed elsewhere", Can I find this value somewhere or filter out to change visualization on history activity from missed call to "call answered elsewhere" ?

Best Regards,

Wesley

slavikbialik:
Hi,
I'm glad to hear that you've been able to sort it out. :)

PR #972, opened 2022-01-21
Add support to use mediaStream on accept invite

thadeu:
This Pull Request allow support to use a custom MediaStream, when accept invite. This will help developers have the flexibility to use the Browser's getUserMedia or not.

How to use? Simple, pass `MediaStream` or `navigator.mediaDevices.getUserMedia` to `sessionDescriptionHandlerOptions`

```js
invitation.accept({
  sessionDescriptionHandlerOptions: {
    mediaStream: new MediaStream(),
    constraints: {
      audio: true,
      video: false
    }
  }
});
```

This is runtime, instead of in the SDH Factory, let me know what you think that it. 

Why? Chrome browser has been blocked when getUserMedia is invoke in background. My app is broken for this.


package.json
```diff
@@ -50,13 +49,13 @@
...
     "karma-webpack": "^5.0.0",
-    "mock-socket": "^9.0.3",
+    "mock-socket": "^9.1.0",
     "pegjs": "^0.10.0",
     "prettier": "2.2.1",
     "ts-loader": "^8.0.18",
     "ts-pegjs": "0.3.1",
     "typescript": "4.2.3",
-    "webpack": "^5.26.0",
+    "webpack": "^5.67.0",
```
thadeu, line 58: This is necessary because LoaderRunner make all spec fails

package.json
```diff
@@ -50,13 +49,13 @@
     "karma-mocha-reporter": "^2.2.5",
     "karma-sourcemap-loader": "^0.3.8",
     "karma-webpack": "^5.0.0",
-    "mock-socket": "^9.0.3",
+    "mock-socket": "^9.1.0",
```
thadeu, line 52: This is necessary in order to import Client from mock-socket https://github.com/onsip/SIP.js/pull/972/files#diff-c8df3bcb89d2bae5907ec3711283b7c2aa02cf080f6c4d610fa61bc8e2075b0bR3

john-e-riordan:
Hi. Thanks for the suggestion.

While the specific case of calling `accept()` arguably lends itself to the caller easily providing a the local media stream in specific applications, there are a number of use cases where the session description handler needs to "callback" for the local media stream when various SIP messages are sent and received to deal with offers and answers - including sending initial INVITEs without SDP (SDP in ACK), responding to in dialog INVITES, and dealing with reliable provisional responses and PRACK. While we arguably could perhaps do better, the session description handler factory and media stream factory is the solution we currently have to "simplify" the circumstances for the developer and make sure things "just work" in all scenarios.

So in short, while your suggestion may make complete sense for your use cases, it falls short as a "standard" or "recommended" or "built in" approach to accepting an invitation because it could confuse and/or lead people to think their use case is covered (it might be, but it also might not be).

All that said, I've taken some, but not all, of your pull request and pushed a modification which will allow you to utilize the approach you are suggesting via a custom media stream factory. 

```
// 1. Create an interface extending SDH options to add mediaStream property.
interface MySessionDescriptionHandlerOptions extends SessionDescriptionHandlerOptions {
  mediaStream: MediaStream;
}

// 2. Implement a custom media stream factory which returns the media stream passed to it via SDH options.
const myMediaStreamFactory: MediaStreamFactory = (
  constraints: MediaStreamConstraints,
  sessionDescriptionHandler: SessionDescriptionHandler,
  options?: SessionDescriptionHandlerOptions
): Promise<MediaStream> => {
  function isMySessionDescriptionHandlerOptions(options?: any): options is MySessionDescriptionHandlerOptions {
    if (options && options.mediaStream instanceof MediaStream) {
      return true;
    }
    return false;
  }
  if (!isMySessionDescriptionHandlerOptions(options)) {
    throw new Error("Options not MySessionDescriptionHandlerOptions.");
  }
  return Promise.resolve(options.mediaStream);
};

// 3. Pass the custom media stream factory as an option when constructing the user agent
userAgentOptions.sessionDescriptionHandlerFactory = defaultSessionDescriptionHandlerFactory(myMediaStreamFactory);

// 4. Add a MediaStream as an option to pass to accept()
const sessionDescriptionHandlerOptions = {
  constraints: {
    audio: true,
    video: false
  },
  mediaStream: new MediaStream()
};

// 5. Call accept with the MediaStream
this.session.accept({ sessionDescriptionHandlerOptions });
```

I believe that is exactly what you are looking to do with just a bit of extra code up front for the custom media stream factory.

It's on [main](https://github.com/onsip/SIP.js/tree/main) here b26885ce24a5788dd3f3db45f726c93e6b7b3763

Will go out in next release.

john-e-riordan:
Also note that all the dependencies you updated got brought up to date in the 0.20.1 release including development support for current versions of node.js.

thadeu:
Thanks @john-e-riordan for your response. I'll test your suggestion and come back if necessary.

PR #988, opened 2022-05-04
Allow +sip.instance in more requests than only REGISTER

danvirsen:
I needed to include instanceId in my `INVITE` requests and could not find a way to do it in the current API.
I shuffled some things around and moved `instanceId` up so that it's always generated and stored in the UA.

There is now a way to add instanceId in all requests that include the contact header.
The instanceId can be automatically generated or manually set.
The old behaviour of only sending instanceId in REGISTER requests is the default.

This introduces a breaking change for people who set their own instanceId on their registerer since `instanceId` is removed from registerer options.

We're using my fork in production right now, but it's a bit of a hassle to handle since the package needs to be built, we can't just point the dependency in package.json to the repository.

So I created this PR to check if this is a change that more people want, and if the implementation looks ok.


john-e-riordan:
Hi,

I've modified your PR...
- an option has been added to `UserAgentOptions` called `instanceId` to specify a custom instance id (default random)
- an option has been added to `UserAgentOptions` called `instanceIdAlwaysAdded` (default false)
- the `instanceId` option that is part of `RegistererOptions` has been deprecated (but otherwise unchanged in behavior)

Furthermore, this PR is now non-breaking and backwards compatible.

It's been picked onto the main branch and will go out with the next release.

Cheers!

danvirsen:
Thanks for taking a look!
My implementation was very much based on our needs, so your changes makes it a lot cleaner and makes more sense.

PR #989, opened 2022-05-05
fix REGISTER GRUU contact

zusrut:
REGISTER request must not use GRUU as contact value. (it breaks RURI and its parameters)

According [rfc5627](https://datatracker.ietf.org/doc/html/rfc5627)

> [3.3](https://datatracker.ietf.org/doc/html/rfc5627#section-3.3).  Using a GRUU
> 
>    Once a user agent obtains GRUUs from the registrar, it uses them in
>    several ways.  First, it uses them as the contents of the Contact
>    header field in non-REGISTER requests and responses that it emits
>    (for example, an INVITE request and 200 OK response).  According to
>    [RFC 3261](https://datatracker.ietf.org/doc/html/rfc3261) [[1](https://datatracker.ietf.org/doc/html/rfc5627#ref-1)], the Contact header field is supposed to contain a URI
>    that routes to that user agent.  Prior to this specification, there
>    hasn't been a way to really meet that requirement.  The user agent
>    would use one of its temporary GRUUs for anonymous calls, and use its
>    public GRUU otherwise.
> 
>    Second, the UA can use the GRUU in any other place it needs to use a
>    URI that resolves to itself, such as a webpage.

### DIFF
--- a/src/api/registerer.ts
+++ b/src/api/registerer.ts
@@ -664,7 +664,7 @@ export class Registerer {
    * Generate Contact Header
    */
   private generateContactHeader(expires: number): string {
-    let contact = this.userAgent.contact.toString();
+    let contact = this.userAgent.contact.toString({ register: true });
     if (this.options.regId && this.options.instanceId) {
       contact += ";reg-id=" + this.options.regId;
       contact += ';+sip.instance="<urn:uuid:' + this.options.instanceId + '>"';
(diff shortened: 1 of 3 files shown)


john-e-riordan:
Thank you for the fix! And including relevant RFC reference very helpful.

Picked onto main branch and added a comment. Will go out in next release.

PR #1013, opened 2022-10-24
session-manager.ts: Optionally send DTMF via SDH.

seanbright:
Obviously the setting name is bad and probably in the wrong place but I wanted a baseline to get a conversation going.

I want to be able to send DTMF via RTP with `SimpleUser` (I really don't want to roll my own version of `SimpleUser` just for this). Thoughts on if this is something you will entertain and, if so, what would you accept in terms of configuration flag/location?

### DIFF
--- a/src/platform/web/session-manager/session-manager-options.ts
+++ b/src/platform/web/session-manager/session-manager-options.ts
@@ -25,6 +25,9 @@ export interface SessionManagerMedia {
 
   /** Local HTML media elements. */
   remote?: SessionManagerMediaRemote | ((session: Session) => SessionManagerMediaRemote);
+
+  /** Whether to use the SessionDescriptionHandler to send DTMF */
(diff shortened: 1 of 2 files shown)


seanbright:
Obviously the setting name is bad and probably in the wrong place but I wanted a baseline to get a conversation going.

I want to be able to send DTMF via RTP with `SimpleUser` (I really don't want to roll my own version of `SimpleUser` just for this). Thoughts on if this is something you will entertain and, if so, what would you accept in terms of configuration flag/location?

john-e-riordan:
Made a few small modifications to naming and so forth, but done and released in 0.21.1. I needed to push a fix to 0.21.0 so added this along with it. Anyway, others have asked for the same, so thanks for getting this started and moving along.

seanbright:
Great! Thanks for including this.

PR #1026, opened 2023-02-07
Support customizing the transport protocol used for Via headers

radioprotector:
This allows for transport protocols that are not based solely on `Transport.server`'s URL scheme.

This provides a path for eliminating the [deprecated `UserAgentOptions.hackViaTcp` configuration option](https://github.com/onsip/SIP.js/blob/b26885ce24a5788dd3f3db45f726c93e6b7b3763/docs/api/sip.js.useragentoptions.hackviatcp.md), as the transport options can now have `{ headerProtocol: "TCP" }` in its place.

This change is also useful when working with SIP endpoints that are behind a load balancer with SSL termination at the balancer. In that case, the transport protocol between the browser and the load balancer would use the SIPS/WSS/TLS protocol, but I believe [the transport protocol indicated in the Via header](https://www.rfc-editor.org/rfc/rfc3261#section-20.42) would need to be SIP/WS/TCP when it reaches the SIP endpoints.

## Potential Alternative

Instead of impacting `Transport.protocol`, I considered adding an alternate `Transport.headerProtocol` property that would be referenced in `setViaHeader` calls.

However, the core `Transport.protocol` property states that it is formatted "as defined for the Via header sent-protocol transport", and `Transport.protocol` is currently only referenced for the purposes of `setViaHeader` calls, so I felt this would be redundant.

### DIFF
--- a/src/platform/web/transport/transport-options.ts
+++ b/src/platform/web/transport/transport-options.ts
@@ -31,4 +31,10 @@ export interface TransportOptions {
    * @defaultValue `true`
    */
   traceSip?: boolean;
+
+  /**
+   * The transport protocol to use in Via header declarations.
+   * If not specified, will be inferred from the server URL.
+   */
+  headerProtocol?: string;
 }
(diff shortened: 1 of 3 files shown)


tal952:
As an hack, you can do: `simpleUser.sessionManager.userAgent._transport._protocol = "WS";`

PR #1094, opened 2024-12-04
Parse scope idetifier on IPv6 addresses

InfinityManu:
During development we required IPv6 with scope identfiiers in the SIP Via and Contact Header. 

New the SIP Via header in the format: 
- Via: ...;received:fe80::XXXX:XXXX:XXXX:XXXX%eth0;....
And the SIP Contact header:
- Contact: <sip:[fe80::XXXX:XXXX:XXXX:XXXX%eth0]:Port;...>
can be parsed.

Threre is no additional Field in the NameAddrHeader -> URI. The whole IPv6 address is still stored in the host property.

### DIFF
--- a/src/grammar/pegjs/src/grammar.pegjs
+++ b/src/grammar/pegjs/src/grammar.pegjs
@@ -177,8 +177,9 @@ IPv6address     = ( h16 ":" h16 ":" h16 ":" h16 ":" h16 ":" h16 ":" ls32
 
 h16             = HEXDIG HEXDIG? HEXDIG? HEXDIG?
 
-ls32            = ( h16 ":" h16 ) / IPv4address
+ls32            = ( h16 ":" h16 (scope_identifier)? ) / IPv4address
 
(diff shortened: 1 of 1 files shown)


seanbright:
Is there an RFC that describes this change?

InfinityManu:
https://www.rfc-editor.org/rfc/rfc4291.html

seanbright:
Is there a SIP related RFC that describes this change?

FWIW, I won’t be able to actually approve or merge this change but changing the grammar without a specification to back up the change would be problematic for interoperability.

InfinityManu:
For anyone wondering this can happen during call setup to asterisk in local IPv6 mode.

seanbright:
Can you post a sample Asterisk config (I am presuming chan_pjsip here and not chan_sip) that I can look at? Nothing in the SIP RFCs indicate that the scope should be part of the signaling, so it may be an Asterisk bug that needs to be addressed.

